PR #8, opened 2016-10-20
DEV: name without dot

licode:
Exclude dot from key names.

### DIFF
--- a/event_model/schemas/event_descriptor.json
+++ b/event_model/schemas/event_descriptor.json
@@ -37,6 +37,11 @@
                     "shape"
                 ],
                 "type": "object"
+            },
+            "data_type": {
+                "title" : "data_type",
+                "patternProperties": {"^([^\.]+)$": {"$ref": "#/definitions/data_type"}},
+                "additionalProperties": false,
             }
         },
         "properties": {
@@ -63,7 +68,7 @@
             }
         },
         "patternProperties": {
-            "^([^\.]+)$": {}},
+            "^([^\.]+)$": {"$ref": "#/definitions/data_type"}},
         "additionalProperties": false,
         "required": [
             "uid",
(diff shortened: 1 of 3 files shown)


licode:
address https://github.com/NSLS-II/bluesky/issues/572

event_model/schemas/event_descriptor.json
```diff
@@ -62,6 +62,9 @@
                 "description": "Creation time of the document as unix epoch time."
             }
         },
+        "patternProperties": {
+            "^([^\.]+)$": { "type": ["object", "integer", "string", "number", "array"] }},
```
licode: yes. it can be empty dict.

danielballan:
Nice!

I am pretty sure we don't want to set additionalProperties to false. Doesn't that disallow user-specified metadata? The event schema does block additional properties (by design) but all of the other documents are "dynamic," allowing arbitrary extra fields.

tacaswell:
I think if you do not set 'additionalProperties' to False, then the format checking (effectively) does not apply.  If I understand it correctly, it is 'may there be properties that are not permitted by some other rule'.

@licode Does this apply recursively? ex, 
```
doc = {...,  # mandatory start doc stuff
       'valid_key': {'invalid.subkey': 1}}
```
should fail.

licode:
Now this can check name recursively.

danielballan:
Nice!

I wrote a test for bluesky verifying that names with dots raise validation errors (when run against this branch of event_model). I found some bugs -- this actually isn't valid JSON. One way to test that is simply to try importing the event_model package.

I opened up a PR against your branch. With those fixes, my bluesky test passes, so this should be good to merge. I'll push my bluesky test once this is merged.

licode:
Neat. I just merged that.

danielballan:
Great. Thanks again for diving into jsonschema, @licode.

PR #18, opened 2018-07-15
[WIP]: Adding docs to event model

jrmlhermitte:
Adding docs to event model


danielballan:
We currently store most of the technical documentation about our data model in NSLS-II/docs. If we move that content here, we can make NSLS-II/docs more like a high-level overview. This is similar to what dask and Jupyter have recently done with their landing pages.

danielballan:
Superceded by #46, which uses our scientific-python-cookiecutter to accomplish the same thing.

PR #19, opened 2018-08-01
MNT: If an EventDescriptor has a 'name' it should be a string.

danielballan:
The 'name' field in EventDescriptors was introduced late and has always been
optional. This change leaves it optional, but requires that if the 'name' field
is present, that it is a string. We make this assumption in a couple places
downstream so it would be good to enforce it here.

### DIFF
--- a/event_model/schemas/event_descriptor.json
+++ b/event_model/schemas/event_descriptor.json
@@ -87,6 +87,10 @@
             },
             "hints": {
                 "$ref": "#/definitions/object_hints"
+            },
+            "name": {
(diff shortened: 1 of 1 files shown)


awalter-bnl:
looks good to me, would merge it but don't yet have write access to event-model !-)

danielballan:
@awalter-bnl Now you do.

Nevertheless I would like @tacaswell eyes on one this before merging, as he was also there for the discussions that introduced 'name' and might spot a pitfall with this change that I have not anticipated.

PR #20, opened 2018-08-03
Typo-fix

elParcero:
FIX: Measurement was spelled measument.


danielballan:
@awalter-bnl You have an invitation to become an Owner of the NSLS-II org. You have to accept the invitation. I just resent it in case I failed to do it correctly last time.

elParcero:
A pleasure @awalter-bnl

PR #24, opened 2018-10-30
ENH: Add document-composing tools.

danielballan:
The only public function is ``create_run``. It returns closures over other
functions, passing internal shared state to manage foreign keys, counters,
and some inter-document validation.

Submitted for early review. Do not merge.

Example usage:

```
import event_model                                                                                                                                                                            
start_doc, create_descriptor, create_resource, create_stop = event_model.create_run()                                                                                                         
print(start_doc)                                                                                                                                                                              
descriptor_doc, create_event = create_descriptor(data_keys={'motor': {'shape': [], 'dtype': 'number'}, 'image': {'shape': [512, 512], 'dtype': 'number', 'external': 'FILESTORE:'}}, name='pri
mary')                                                                                                                                                                                        
print(descriptor_doc)                                                                                                                                                                         
resource_doc, create_datum = create_resource(spec='TIFF', root='/tmp', resource_path='stack.tiff', resource_kwargs={})                                                                        
print(resource_doc)                                                                                                                                                                           
datum_doc = create_datum(datum_kwargs={'slice': 5})                                                                                                                                           
print(datum_doc)                                                                                                                                                                              
event_doc = create_event(data={'motor': 0, 'image': datum_doc['datum_id']}, timestamps={'motor': 0, 'image': 0}, filled={'image': False}, seq_num=1)                                          
print(event_doc)                                                                                                                                                                              
stop_doc = create_stop()                                                                                                                                                                      
print(stop_doc)    
```

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -10,7 +10,7 @@ import time as ttime
 import uuid
 from ._version import get_versions
 
-__all__ = ['DocumentNames', 'schemas', 'create_run']
+__all__ = ['DocumentNames', 'schemas', 'compose_run']
 
 
 class DocumentNames(Enum):
@@ -47,16 +47,16 @@ for name, filename in SCHEMA_NAMES.items():
 __version__ = get_versions()['version']
 del get_versions
 
-CreateRunBundle = namedtuple('CreateRunBundle',
-                             'start_doc create_desriptor create_resource create_stop')
-CreateDescriptorBundle = namedtuple('CreateDescriptorBundle',
-                                    'descriptor_doc create_event')
-CreateResourceBundle = namedtuple('CreateResourceBundle',
-                                  'resource_doc create_datum')
+ComposeRunBundle = namedtuple('ComposeRunBundle',
+                             'start_doc compose_descriptor compose_resource compose_stop')
+ComposeDescriptorBundle = namedtuple('ComposeDescriptorBundle',
+                                    'descriptor_doc compose_event')
+ComposeResourceBundle = namedtuple('ComposeResourceBundle',
+                                  'resource_doc compose_datum')
 
 
 
-def create_datum(*, resource, counter, datum_kwargs, validate=True):
+def compose_datum(*, resource, counter, datum_kwargs, validate=True):
     resource_uid = resource['uid']
     doc = {'resource': resource_uid,
            'datum_kwargs': datum_kwargs,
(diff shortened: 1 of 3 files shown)


danielballan:
Note: I'm passing the whole document down into the closure for now, here and elsewhere. Eventually I think it would be better to pass down only the parts we need, for both safely (avoid shared mutable state where possible) and ability to reuse pieces of this API independently. For example, ophyd will want to run `create_resource` and `create_datum` without having run `create_run` or having access to a `start` document.

ronpandolfi:
This is great. Could we also have a context-managed equivalent of `create_run` that generates `stop_doc` and runs a callback?

tacaswell:
Something like

``python
with create_run as bundle_of_stuff:
    ...
```

Having that for all of the generations makes sense to me too...

Having `create_run` return a named tuple or SimpleNamespace would make sense to me.

I am partial to defined in-line functions, but partial makes more sense incase we want those low-level functions else where?

danielballan:
> Could we also have a context-managed equivalent of create_run that generates stop_doc and runs a callback?

> Having that for all of the generations makes sense to me too...

Sure, I'll try building context managers on top of the functions and see how it looks.

> Having create_run return a named tuple or SimpleNamespace would make sense to me.

Yeah, I thought about a `namedtuple`. Will do.

> I am partial to defined in-line functions, but partial makes more sense incase we want those low-level functions else where?

Right. I started with in-line functions and quickly changed course for that reason. For example, ophyd will need to be able to use `create_resource` without knowing the RunStart yet, maybe putting in `{'run_start': None}` as a placeholder. It might also be useful to be able to access these things individually in tests, I guess, but it's not that big a deal to just generate them. Meh.

CJ-Wright:
:+1:

tacaswell:
> For example, ophyd will need to be able to use `create_resource` without

maybe the Resource factory should just be stand alone them?  How terrible would it look to pass a resource factory into `trigger` (I think pretty terrible, but could be convinced otherwise).

danielballan:
What's the rationale for defining functions locally? It makes them impossible to individually document and test. It seems a bit too clever to my taste.

If the goal is to make sure they are definitely not part of the public API, I think leaving them out of `__all__` is good enough for now. Is there another reason?

danielballan:
I would like to propose merging this in its current form and experimenting with context managers separately, aiming to eventually move experiments up stream when we find a good pattern.

awalter-bnl:
merging as it has been approved and I would like to use it for suitcase_csv tests

PR #27, opened 2018-12-10
MNT: Add name for demensions as dims to data_key

ke-zhang-rd:
For [issue26](https://github.com/NSLS-II/event-model/issues/26)

### DIFF
--- a/event_model/schemas/event_descriptor.json
+++ b/event_model/schemas/event_descriptor.json
@@ -27,6 +27,13 @@
                         },
                         "description": "The shape of the data.  Null and empty list mean scalar data."
                     },
+                    "dims": {
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        },
+                        "description": "The names for dimensions of the data. Null and empty list if scalar data"
+                    },
                     "source": {
                         "type": "string",
                         "description": "The source (ex piece of hardware) of the data."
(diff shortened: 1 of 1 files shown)


event_model/schemas/event_descriptor.json
```diff
@@ -27,6 +27,13 @@
                         },
                         "description": "The shape of the data.  Null and empty list mean scalar data."
                     },
+                    "dims": {
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        },
+                        "description": "The names for dimensions of the data. Null and empty list if scalar data"
```
danielballan: I think "Null _or_ empty list if scalar data" would make more sense than "and" here.

ke-zhang-rd:
Closes#26

danielballan:
:+1:

event_model/schemas/event_descriptor.json
```diff
@@ -27,6 +27,13 @@
                         },
                         "description": "The shape of the data.  Null and empty list mean scalar data."
                     },
+                    "dims": {
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        },
+                        "description": "The names for dimensions of the data. Null or empty list if scalar data"
```
tacaswell: Is there a way to express in the schema that if this is non-null this must be the same length as shape?

danielballan: I could not find one. We can enforce it in the composing tools at least.

PR #28, opened 2018-12-13
Fix document-composing functions

danielballan:
In #24 I completely misused `jsonschema.validate` and was not actually validating the documents. This PR employs actual validation and fixes a couple important bugs revealed by validation in the tests.


tacaswell:
womp womp.  Sorry I missed this on review.

PR #29, opened 2018-12-14
ENH: Add event_page, datum_page, DocumentRouter.

danielballan:
This is a sequel to `bluesky.callbacks.CallbackBase`.

Improvements:
* By putting it in event-model, packages like suitcase-csv can use it without importing bluesky as a run dependency.
* It returns things, which makes it useful in pipelines.
* We can take this opportunity to adopt a name that correctly suggests that its usefulness isn't confined to callbacks.

I have also sketched out `event_page` (seems like the consensus choice?) and `datum_page` (still contentious) ~as aliases for `bulk_event` and `bulk_datum`~. A `bulk_event` and `bulk_datum` uses a "list of dicts" representation. An `event_page` and `datum_page` use a "dict of lists" DataFrame-like representation. A have added schemas for the new `event_page` and `datum_page` concepts and, belatedly, added a schema for `bulk_datum`. Both of the `bulk_*` representations are deprecated.

<hr />

Edit: I should be sure to credit @CJ-Wright for advocating for the second bullet point above in https://github.com/NSLS-II/bluesky/pull/1122.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -46,7 +46,7 @@ class DocumentRouter:
 
         (name, getattr(router, name)(doc))
     """
-    def __call__(self, name, doc):
+    def __call__(self, name, doc, validate=False):
         """
         Process a document.
 
@@ -54,14 +54,22 @@ class DocumentRouter:
         ----------
         name : string
         doc : dict
+        validate : boolean
+            Apply jsonschema validation to the documents coming *out*. This is
+            False by default.
 
         Returns
         -------
-        name, new_doc : string, dict
+        name, output_doc : string, dict
             The same name as what was passed in, and a doc that may be the same
             instance as doc, a copy of doc, or a different dict altogether.
         """
-        return name, getattr(self, name)(doc)
+        ret = getattr(self, name)(doc)
+        output_doc = getattr(self, name)(doc)
+        if validate:
+            jsonschema.validate(output_doc,
+                                schemas[getattr(DocumentNames, name)])
+        return (name, output_doc) if output_doc is not None else output_doc
 
     def start(self, doc):
         return doc
@@ -76,10 +84,16 @@ class DocumentRouter:
         return doc
 
     def event(self, doc):
-        return doc
+        event_page = pack_event_into_event_page(doc)
+        output_event_page = self.event_page(event_page)
+        output_event = unpack_event_page_into_event(output_event_page)
+        return output_event
 
     def datum(self, doc):
-        return doc
+        datum_page = pack_datum_into_datum_page(doc)
+        output_datum_page = self.datum_page(datum_page)
+        output_datum = unpack_datum_page_into_datum(output_datum_page)
+        return output_datum
 
     def event_page(self, doc):
         return doc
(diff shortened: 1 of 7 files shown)


CJ-Wright:
Should the `RunRouter` also go here?

event_model/__init__.py
```diff
@@ -25,6 +26,82 @@ class DocumentNames(Enum):
...
+        return self.event_page(doc)
+
+    def bulk_datum(self, doc):
+        # Do not modify this in a subclass. Use event_page.
+        warnings.warn(
+            "The name 'bulk_datum' has been deprecated in favor of "
+            "'datum_page'.")
+        return self.datum_page(doc)
+
+
```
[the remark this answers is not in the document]
danielballan: Exactly. Should help mitigate confusion about what it means.

danielballan:
Yes, @CJ-Wright, that sounds right.

event_model/__init__.py
```diff
@@ -25,6 +26,82 @@ class DocumentNames(Enum):
...
+    def datum_page(self, doc):
+        return doc
+
+    def bulk_event(self, doc):
+        # Do not modify this in a subclass. Use event_page.
+        warnings.warn(
+            "The name 'bulk_event' has been deprecated in favor of "
+            "'event_page'.")
+        return self.event_page(doc)
```
tacaswell: This will need to transpose the input.  Currently `bulk_events` are `list of dicts` where as `event_page` is going to duck-type as a dataframe so should be a `dict of lists`.

event_model/__init__.py
```diff
@@ -25,6 +26,82 @@ class DocumentNames(Enum):
...
+
+    def bulk_datum(self, doc):
+        # Do not modify this in a subclass. Use event_page.
+        warnings.warn(
+            "The name 'bulk_datum' has been deprecated in favor of "
+            "'datum_page'.")
+        return self.datum_page(doc)
```
tacaswell: same as `bulk_event`, needs transpose.

event_model/__init__.py
```diff
@@ -25,6 +26,82 @@ class DocumentNames(Enum):
...
+        name : string
+        doc : dict
+
+        Returns
+        -------
+        name, new_doc : string, dict
+            The same name as what was passed in, and a doc that may be the same
+            instance as doc, a copy of doc, or a different dict altogether.
+        """
+        return name, getattr(self, name)(doc)
```
tacaswell: Suggest changing to:

```suggestion
         ret = getattr(self, name)(doc)
         return name, ret if ret is not None else doc
```

so that if the down-stream implementer forgets to return it will not break everything.

CJ-Wright:
How are you planning on enforcing this?

danielballan:
We can employ "administrative protections" to nudge people toward the usage we recommend. We'll follow it in our own examples (in the new callbacks in bluesky, for example), we'll document it, we'll ask our collaborators to follow it. Also, we'll avoid laying track for usages that are out of scope.

danielballan:
This is ready for another round of review.

tacaswell:
It may also be worth thinking about how to 'promote' things that we are collecting as Events -> EventPages (ala https://github.com/NSLS-II/databroker/blob/master/databroker/pivot.py).

danielballan:
Isn't that what `pack_event_page(*events)` does?

event_model/schemas/bulk_datum.json
```diff
@@ -0,0 +1,27 @@
+{
+    "properties": {
+        "datum_kwargs_list": {
```
tacaswell: It is `datum_kwarg_list` https://github.com/NSLS-II/databroker/blob/bdfef404ede4eb94f78dec8769ac3d7c675b4484/databroker/assets/base_registry.py#L639-L643

tacaswell:
> Isn't that what `pack_event_page(*events)` does?

No, this is a single event which has lists as the values -> an event page of many events.

Good except for a spelling error

PR #33, opened 2019-01-09
ENH: Add Filler.

danielballan:
This is a working sketch that I live-coded at ISS for our "Minimum Viable Product" analysis pipeline demo.

The retry/backoff logic in `event` uses `time.sleep(...)`. We may want to refactor that so that it can be plugged into a sync, threaded, or async framework.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -127,8 +127,8 @@ class Filler(DocumentRouter):
     Filler manages caches of potentially expensive resources (e.g. large data
     in memory) managing its lifecycle is important. If used as a context
     manager, it will drop references to its caches upon exit from the
-    context. Unless the user holds additionally references to those caches,
-    they will be garbage collected.
+    context. Unless the user holds additional references to those caches, they
+    will be garbage collected.
 
     But for some applications, such as taking multiple passes over the same
     data, it may be useful to keep a longer-lived Filler instance and then
@@ -151,9 +151,10 @@ class Filler(DocumentRouter):
 
             handler_instance(**datum_kwargs)
 
-        As implied by the names, this is typically implemented using a class that
-        implements ``__init__`` and ``__call__``, with the respective signatures.
-        In general it may be any callable-that-returns-a-callable.
+        As the names 'handler class' and 'handler instance' suggest, this is
+        typically implemented using a class that implements ``__init__`` and
+        ``__call__``, with the respective signatures. But in general it may be
+        any callable-that-returns-a-callable.
     include : Iterable
         The set of fields to fill. By default all unfilled fields are filled.
         This parameter is mutually incompatible with the ``exclude`` parameter.
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -213,6 +213,7 @@ class Filler(DocumentRouter):
         self.retry_intervals = list(retry_intervals)
         self.include = include
         self.exclude = exclude
+        self._closed = False
 
     @staticmethod
     def get_default_datum_cache():
@@ -306,15 +307,25 @@ class Filler(DocumentRouter):
         # them it's the user's problem to manage their lifecycle. If the user
         # does not (e.g. they are the default caches) the gc will look after
         # them.
+        self._closed = True
         self._handler_cache = None
         self._datum_cache = None
 
+    def __call__(self, name, doc, validate=False):
+        if self._closed:
+            raise EventModelRuntimeError(
+                "This Filler has been closed and is no longer usable.")
+        super().__call__(name, doc, validate)
 
 class EventModelError(Exception):
     ...
 
 
 class EventModelValueError(EventModelError, ValueError):
+    ...
+
+
+class EventModelRuntimeError(EventModelError, RuntimeError):
     ...
 
 
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -145,12 +145,12 @@ class Filler(DocumentRouter):
         A cache mapping a Resource uid to a handler instance created for that
         resource. That instance may hold on to certain expensive resources,
         such as data in memory or an open file handle. If None, a dict is used
-        and cleared when :meth:`start` or :meth:`stop` are called. If a cache
-        is provided by the user, the user is in charge of clearing it.
+        and cleared on ``__exit__`` from the context. If a cache is provided by
+        the user, the user is in charge of clearing it.
     datum_cache : dict, optional
         A cache mapping a Datum id to a Datum document. If None, a dict is used
-        and cleared when :meth:`start` or :meth:`stop` are called. If a cache
-        is provided by the user, the user is in charge of clearing it.
+        and cleared on ``__exit__`` from the context. If a cache is provided by
+        the user, the user is in charge of clearing it.
     retry_intervals : Iterable, optional
         If data is not found on the first try, there may a race between the
         I/O systems creating the external data and this stream of Documents
@@ -178,7 +178,6 @@ class Filler(DocumentRouter):
         self.retry_intervals = list(retry_intervals)
 
     def start(self, doc):
-        self._auto_clear()
         return doc
 
     def resource(self, doc):
@@ -226,14 +225,12 @@ class Filler(DocumentRouter):
         return doc
 
     def stop(self, doc):
-        self._auto_clear()
         return doc
 
-    def _auto_clear(self):
-        # As documented in the Parameters section, the caches are cleared
-        # automatically upon start and stop if the default caches are used. If
-        # the user passes in a custom cache, they control the clearing
-        # behavior.
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *exc_details):
         if self._auto_clear_handler_cache:
             self._handler_cache.clear()
         if self._auto_clear_datum_cache:
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -284,7 +284,7 @@ class Filler(DocumentRouter):
                         actual_data = handler(**datum_doc['datum_kwargs'])
                         doc['data'][key] = actual_data
                         doc['filled'][key] = datum_id
-                    except OSError as error_:
+                    except IOError as error_:
                         # The file may not be visible on the network yet.
                         # Wait and try again. Stash the error in a variable
                         # that we can access later if we run out of attempts.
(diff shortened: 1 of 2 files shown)


event_model/__init__.py
```diff
@@ -119,6 +119,74 @@ def bulk_datum(self, doc):
         self.datum_page(bulk_datum_to_datum_page(doc))
 
 
+
+class Filler:
+    """Pass documents through, loading any externally-referenced data."""
+    ATTEMPTS = 10
+
+    def __init__(self, handler_registry, handler_cache=None, datum_cache=None):
```
tacaswell: :+1: for passing the caches in.

Would advocate for kwarg-only, but I think we still have to support py2 on this repo?

danielballan:
I'm not sure how close this is to what we need, but I would like some component in event-model that can be used by databroker, `bluesky.callbacks` (like finally getting an official image viewer done), suitcase (e.g. TIFF export attn @awalter-bnl), and third-party data processing pipelines.

event_model/__init__.py
danielballan: Actually we do not! Some time over the summer @stuartcampbell gave his blessing for py2-only databroker. And this would not preclude writing a py2-compat data access client anyway. Keyword-only it is....

tacaswell:
I have my standing concerns about managing the life-cycle / resources of the handlers.

Maybe `stop` should remove any handlers it created?

Maybe the default cache should be capped (which means we also need to cache the resource documents)?

danielballan:
> Maybe stop should remove any handlers it created?

I envisioned leaving that up to the subclass. One-run callbacks wouldn't need to worry about it because they would just be wholly garbage collected. Callbacks with larger scope can roll their own management, either something easy like:

```py
def stop(self):
    self.handlers.clear()
```

or something more sophisticated make reference to the size of the loaded data.

But if in your judgement this is likely to create a bunch of memory-killing code on the floor, I would accept hard-coding the clear-handlers-cache-on-stop approach, at least as a first pass.

event_model/__init__.py
```diff
@@ -119,6 +119,74 @@ def bulk_datum(self, doc):
...
+    def event(self, doc):
+        for key, is_filled in doc['filled'].items():
+            if not is_filled:
+                datum_id = doc['data'][key]
+                datum_doc = self.datums[datum_id]
+                interval = 0.001
+                error = None
+                for i in range(self.ATTEMPTS):
+                    try:
+                        handler = self.handlers[datum_doc['resource']]
```
CJ-Wright: It might be nice to have this as an independent method. That way we can hand in datum ids and get back data.

CJ-Wright: https://github.com/xpdAcq/xpdAn/blob/master/xpdan/vend/callbacks/core.py#L561

event_model/__init__.py
```diff
@@ -119,6 +119,74 @@ def bulk_datum(self, doc):
...
+        self.handler_registry = handler_registry
+        if handler_cache is None:
+            handler_cache = {}
+        if datum_cache is None:
+            datum_cache = {}
+        self.handlers = handler_cache
+        self.datums = datum_cache
+
+    def __call__(self, name, doc):
+        return name, getattr(self, name)(doc)
```
CJ-Wright: This either needs to have a stop method or more magic for if a doc is not returned

tacaswell: Ha!  I had the same thought, wrote the comment, and then deleted it before posting :-p

danielballan:
Points from group discussion:
- If the user passes in a cache, they own the code for clearing it. If not, Filler uses some default code for clearing.
- Call 'clear' on 'start' and 'stop'.

CJ-Wright:
Do we want/need a tool to create the handler registry? Similar to how databrokers auto register handlers.

danielballan:
> Do we want/need a tool to create the handler registry?

Good suggestion. Yes, but that's a configuration management activity best handled by higher layers, I think. For example, databroker looks for `/etc/databroker/...` to register handlers, and I wouldn't want to move that here.

danielballan:
The cache handling has evolved somewhat from where things stood in the group meeting. The current behavior is simpler and easier to explain, and I believe it finds a good balance of "helpful" and "avoids unintended consequences." Developed in conversation with @tacaswell. See docstrings for details.

event_model/__init__.py
```diff
@@ -120,6 +120,176 @@ def bulk_datum(self, doc):
         self.datum_page(bulk_datum_to_datum_page(doc))
 
 
+class Filler(DocumentRouter):
+    """Pass documents through, loading any externally-referenced data.
+
+    It is recommended to use the Filler as a context manager.  Because the
+    Filler manages caches of potentially expensive resources (e.g. large data
+    in memory) managing its lifecycle is important. If used as a context
+    manager, it will drops it references to its caches upon exit from the
```
awalter-bnl:
```suggestion
    manager, it will drop references to its caches upon exit from the
```

event_model/__init__.py
```diff
@@ -120,6 +120,176 @@ def bulk_datum(self, doc):
...
+        In general it may be any callable-that-returns-a-callable.
+    handler_cache : dict, optional
+        A cache of handler instances. If None, a dict is used.
+    datum_cache : dict, optional
+        A cache of Datum documents. If None, a dict is used.
+    retry_intervals : Iterable, optional
+        If data is not found on the first try, there may a race between the
+        I/O systems creating the external data and this stream of Documents
+        that reference it. If Filler encounters an ``IOError`` it will wait a
+        bit and retry. This list specifies how long tosleep (in seconds)
```
awalter-bnl:
```suggestion
        bit and retry. This list specifies how long to sleep (in seconds)
```

event_model/__init__.py
```diff
@@ -120,6 +120,176 @@ def bulk_datum(self, doc):
...
+    ...
+    >>> del filler  # Free up memory from potentially large caches.
+    """
+    def __init__(self, handler_registry, *,
+                 handler_cache=None, datum_cache=None, retry_intervals=None):
+        self.handler_registry = handler_registry
+        if handler_cache is None:
+            handler_cache = self.get_default_handler_cache()
+        if datum_cache is None:
+            datum_cache = self.get_dfeault_datum_cache()
```
awalter-bnl:
```suggestion
            datum_cache = self.get_default_datum_cache()
```

event_model/__init__.py
```diff
@@ -120,6 +120,176 @@ def bulk_datum(self, doc):
...
+        self._handler_cache = handler_cache
+        self._datum_cache = datum_cache
+        if retry_intervals is None:
+            # Total wait of about 2 seconds before giving up.
```
awalter-bnl:
```suggestion
            # Total wait about 2 seconds before giving up.
```

danielballan:
Thanks for the careful read, @awalter-bnl.

We should definitely also add tests before this is merged. But before I go to the trouble I'd like to have consensus on the contours of the design since this has been a fast-moving target up to this point.

awalter-bnl:
Sorry @danielballan for jumping the gun, you did change the WIP tage to ENH which I assumed meant it was ready to go !-). 

In that case, I am happy with the current approach I can't think of any use cases it doesn't cover.

event_model/__init__.py
```diff
@@ -120,6 +120,176 @@ def bulk_datum(self, doc):
...
+    Or as a long-lived object.
+
+    >>> f = Filler(handler_registry)
+    >>> for name, doc in stream:
+    ...     filler(name, doc)
+    ...
+    >>> del filler  # Free up memory from potentially large caches.
+    """
+    def __init__(self, handler_registry, *,
+                 handler_cache=None, datum_cache=None, retry_intervals=None):
```
CJ-Wright: Do we want to enable filling only certain fields?

danielballan: Good point! I think we do.

danielballan:
I addressed the changes discussed today and added a basic test. I think this is ready to go.

event_model/__init__.py
```diff
@@ -120,14 +120,250 @@ def bulk_datum(self, doc):
         self.datum_page(bulk_datum_to_datum_page(doc))
 
 
+# In total wait about 2 seconds before giving up.
+# Max sleep, between the final two attempts, is about 1 second.
+DEFAULT_RETRY_INTERVALS = tuple(0.001 * numpy.array([2**i for i in range(11)]))
```
tacaswell: I see why this is here and that it is too long to put directly in the signature, but this makes me a little sad (it is still not clear what the default retry behavior is from just looking at the signature).  I was thinking like `(.1, .5, 1)`

event_model/__init__.py
```diff
@@ -120,14 +120,250 @@ def bulk_datum(self, doc):
...
+
+    Or as a long-lived object.
+
+    >>> f = Filler(handler_registry)
+    >>> for name, doc in stream:
+    ...     filler(name, doc)
+    ...
+    >>> del filler  # Free up memory from potentially large caches.
+    """
+    def __init__(self, handler_registry, *,
```
tacaswell: This should also take `root_map` as an input.

event_model/__init__.py
```diff
@@ -120,14 +120,250 @@ def bulk_datum(self, doc):
...
+        If an IOError is raised when loading the data after the configured
+        number of attempts. See the ``retry_intervals`` parameter for details.
+
+    Examples
+    --------
+    A Filler may be used as a context manager.
+
+    >>> with Filler(handler_registry) as filler:
+    ...     for name, doc in stream:
+    ...         filler(name, doc)
```
tacaswell: should probably show captureing the events for later analysis?  As written this will fill the events and then let them be GC'd...

event_model/__init__.py
```diff
@@ -120,14 +120,250 @@ def bulk_datum(self, doc):
...
+
+    def resource(self, doc):
+        try:
+            handler_class = self.handler_registry[doc['spec']]
+        except KeyError as err:
+            raise UndefinedAssetSpecification(
+                f"Resource document with uid {doc['uid']} "
+                f"refers to spec {doc['spec']!r} which is "
+                f"not defined in the Filler's handler registry.") from err
+        handler = handler_class(doc['resource_path'],
```
tacaswell: do we want to defer generating the handler instance until we actually need it?

event_model/__init__.py
```diff
@@ -186,7 +432,7 @@ def compose_resource(*, start, spec, root, resource_path, resource_kwargs,
            'spec': spec,
            'root': root,
            'resource_path': resource_path,
-           'resource_kwargs': {},
+           'resource_kwargs': resource_kwargs,
```
tacaswell: good catch!

event_model/__init__.py
danielballan: Done.

event_model/__init__.py
danielballan: Reasonable. Done.

PR #34, opened 2019-01-09
FIX: return type stability on __call__

tacaswell:


danielballan:
That was a typo. Thanks.

PR #35, opened 2019-01-09
FIX: Accept {'filled': <boolean>} in event_page schema.

danielballan:
Per page, I think a given data_key should either be entirely filled or entirely
unfilled.  So, an unfilled page would look like:

```py
{'data': {'image': [<datum1>, <datum2>]}, 'filled': {'image': False}}
```

and a filled page would look like:

```py
{'data': {'image': [<array1>, <array2>]}, 'filled': {'image': [datum_id1, datum_id2]}}
```

This would also be valid, as we still have some wiggle room at the schema
level around the contents of "filled".

```py
{'data': {'image': [<array1>, <array2>]}, 'filled': {'image': True}}
```

Currently the event_page schema *requires* filled to be a object of *arrays*
and thus does not accept ``'filled': {'image': False}`` because ``False`` is
not an array. This PR fixes that and updates the description of filled.

### DIFF
--- a/event_model/schemas/event.json
+++ b/event_model/schemas/event.json
@@ -11,7 +11,7 @@
         "filled": {
             "type": "object",
             "additionalProperties": {"type": ["boolean", "string"]},
-            "description": "Mapping the keys of externally-stored data to a boolean indicating whether that data has yet been loaded"
+            "description": "Mapping each of the keys of externally-stored data to the boolean False, indicating that the data has not been loaded, or to foreign keys (moved here from 'data' when the data was loaded)"
         },
         "descriptor": {
             "type": "string",
--- a/event_model/schemas/event_page.json
+++ b/event_model/schemas/event_page.json
@@ -26,7 +26,7 @@
         },
         "filled": {
             "type": "dataframe_for_filled",
-            "description": "Mapping the keys of externally-stored data to a boolean indicating whether that data has yet been loaded"
+            "description": "Mapping each of the keys of externally-stored data to an array containing the boolean False, indicating that the data has not been loaded, or to foreign keys (moved here from 'data' when the data was loaded)"
         },
         "seq_num": {
             "type": "array",


danielballan:
If this change seems sensible, I should also update the conversion functions to to create a page with `{'filled': False}` rather than `{'filled': [False, False, False, ..., False]}` before this PR is merged.

awalter-bnl:
This looks good to me, and allows for an easy check to see if an event_page has unfilled entries (as per a discussion with @danielballan ).

`if all(event_page['filled'].values()):`

is false if there us unfilled data.

danielballan:
Thanks for taking a look, @awalter-bnl. I'm glad you helped us catch this one early.

@tacaswell, this one should get your eyes also. I think we are on the same "page" (get it?) on this one but I want to be sure.

tacaswell:
> I think we are on the same "page" (get it?)

It is too late to revoke you dad-joke license right?

event_model/schemas/event.json
```diff
@@ -11,7 +11,7 @@
         "filled": {
             "type": "object",
             "additionalProperties": {"type": ["boolean", "string"]},
-            "description": "Mapping the keys of externally-stored data to a boolean indicating whether that data has yet been loaded"
+            "description": "Mapping the keys of externally-stored data to a False, indicating that they have not been loaded, or True, or the foreign keys moved from data, which should also be interpreted as True."
```
tacaswell: I don't understand the third 'or' clause.

tacaswell:
Yes, I agree with the change, a half-filled page would be very annoying to work with.

The flexibility to `'filled': {'image': True}}` is not great but ok.

Is

```
{'data': {'image': [<datum1>, <datum2>]}, 'filled': {'image': [False, False]}}
```

still allowed?

One draw back of this is that it makes the code for consuming this have to deal with this a bit complex (particularly coupled with the `bool(arr)` issues with numpy arrays) 

```python

def is_my_page_filled(page, col):
    try:
        return bool(page[col]['filled'][0])
    except TypeError:
        return bool(page[col]['filled'])
```

which isn't than bad now that I type it out...

awalter-bnl:
> > I think we are on the same "page" (get it?)
> 
> It is too late to revoke you dad-joke license right?

He has just become a Dad, doesn't that give him the write to make dad jokes !-)

tacaswell:
I hear you on the 'json schema can not specify no mixed bool vectors', however we have 3 things in the event that have matching (or at least strict sub-set) keys (data, timestamps, filled).  It seems odd that the `List[Event] <-> EventPage` translations can not be handled the same for all three.  It gets odder than in some cases (when it has been filled and we are tracking the filled state) they do all transform the same, but in other cases (the unfilled or filled but dropping uids case) they do not <insert tensor joke>.

I have been noodling over the Event vs EventPage duality.  I am reverting to the view tho Event is the 'fundemental' atomic representation of the data and EventPage is an efficiency-driven implementation detail for when you want to work with many events.   The motivating things here are EventPages do not have their own UIDs and do not have fixed size (the same dataset chopped up into 5 100 Event EventPages is the same dataset as when in is 100 5 Event EventPages, however if you were to drop / change any single Event, it is no longer the same dataset).

tacaswell:
As discussed in the staff meeting:
 - may have good reason for a mixed-filled Page
 - would like a mulit-level column dataframe to just be a valid EventPage

Which leans towards only having vector values (or at least allowing vectors of bools).

danielballan:
What we decided above (in the group meeting) amounts to declaring the current code to be the desired code, and reduces this PR to documentation change, only touching the `'description'` fields.

PR #36, opened 2019-01-15
Adding `verify_filled` and `santize_doc` functions.

awalter-bnl:
This adds the functions:
1.  `verify_filled`, which takes in an event_page , raises an exception if any of the data is unfilled or returns true if it is filled.
2. `sanitize_doc`, which takes in an event model document and converts any numpy objects to built-in python objects compatible with mongoDB and json.


event_model/__init__.py
```diff
@@ -439,3 +440,113 @@ def _transpose_dict_of_lists(dict_of_lists):
...
+    if isinstance(val, (numpy.generic, numpy.ndarray)):
+        if numpy.isscalar(val):
+            return val.item()
+        return val.tolist()
+    return val
+
+
+def _apply_to_dict_recursively(dictionary, func):
+    '''
+    step through a dictionary of dictionaries recursively and apply a function
```
danielballan: Needs to be a one-liner.

event_model/__init__.py
```diff
@@ -439,3 +440,113 @@ def _transpose_dict_of_lists(dict_of_lists):
...
+    else:
+        verified = True
+        return verified
+
+
+def sanitize_doc(doc):
+    '''Takes in an event-model document and returns a copy which has all the
+    numpy objects converted to built-in pyton versions.
+
+    This is useful for sanitzing documents prior to sending to a mongo database
```
danielballan: It might be worth spelling this out a bit more for people: "prior to sending to any consumer that does not recognize numpy types, such a MongoDB or a JSON encoder".

event_model/__init__.py
```diff
@@ -439,3 +440,113 @@ def _transpose_dict_of_lists(dict_of_lists):
...
+
+    return sanitized_doc
+
+
+def _sanitize_numpy(val):
+    '''Convert any numpy objects into built-in Python types.
+
+    Parameters
+    ----------
+    val : The potential numpy object to be converted.
```
danielballan: The expression after `:` should be a data type, not a description. This can be written as:

```rst
val : object
    The potential numpy object to be converted.
```

event_model/__init__.py
```diff
@@ -439,3 +440,113 @@ def _transpose_dict_of_lists(dict_of_lists):
...
+
+    Returns
+    -------
+    val : The input parameter, converted to a built-in python type if it is a
```
danielballan: Likewise

event_model/__init__.py
```diff
@@ -439,3 +440,113 @@ def _transpose_dict_of_lists(dict_of_lists):
...
+        A function to apply to each value in dictionary.
+    '''
+
+    for key, val in dictionary.items():
+        if hasattr(val, 'items'):
+            dictionary[key] = _apply_to_dict_recursively(val, func)
+        dictionary[key] = func(val)
+
+
+class UnfilledData(Exception):
```
danielballan: We have adopted* the convention that every exception _directly_ raised by a library (i.e. typed out as `raise ___`, not caused indirectly) should inherit from some base exception defined in that library. In the case of this library, every error raised by event_model is some subclass of a `EventModelError`. This helps users see which layer a given error is coming from.

Some exceptions are simple variants of Python's built-in errors. For example, instead of raising `KeyError`, we raise `EventModelKeyError` which is a subclass of both `KeyError` and `EventModelError`. Therefore, it can be "caught" via `except KeyError` just like a normal `KeyError` would be, but it also signals that it is coming from event-model.

We can also have special errors that have no built-in counterpart. There is no examples of this in event-model yet, but there are some in caproto, such as `RemoteProtocolError`. It inherits from `CaprotoError`, as even error raised by caproto does, but does not inherit from anything else.

I think this is in that latter category.

*This is a relatively recent (1-2 year) practice for us. We do it fully consistently in caproto, inconsistently in ophyd, and hardly at all yet in bluesky or databroker.

event_model/__init__.py
```diff
@@ -439,3 +440,113 @@ def _transpose_dict_of_lists(dict_of_lists):
...
+
+def verify_filled(event_page):
+    '''Take an event_page document and verify that it is completley filled
+
+    Parameters
+    ----------
+    event_page : event_page document
+        The event page document to check
+
+    Raise
```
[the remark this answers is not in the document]
awalter-bnl: fixed

event_model/__init__.py
```diff
@@ -439,3 +440,113 @@ def _transpose_dict_of_lists(dict_of_lists):
...
+    -----
+    UnfilledData : exception
+        Raised if any of the data in the event_page is unfilled, when raised it
+        inlcudes a list of unfilled data objects in the exception message.
+
+    Returns
+    -------
+    verified : Boolean
+        To indicate if it is correctly filled, this is done make code easier to
+        read (i.e. use `if verified_filled(event_page) ...`) it will never
```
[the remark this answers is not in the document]
awalter-bnl: changed

event_model/__init__.py
```diff
@@ -439,3 +440,113 @@ def _transpose_dict_of_lists(dict_of_lists):
...
+                                           "{!r}. Try passing the parameter "
+                                           "`gen` through `event_model.Filler`"
+                                           " first.".format(unfilled_data))
+    else:
+        verified = True
+        return verified
+
+
+def sanitize_doc(doc):
+    '''Takes in an event-model document and returns a copy which has all the
```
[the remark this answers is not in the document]
awalter-bnl: fixed

event_model/__init__.py
awalter-bnl: fixed

event_model/__init__.py
awalter-bnl: fixed

event_model/__init__.py
awalter-bnl: fixed

event_model/__init__.py
awalter-bnl: fixed

event_model/__init__.py
awalter-bnl: I am for this standard, but once again raise the fact that word of mouth is not a useful or adequate way of passing this information to new contributors. For a projects with a large, multi-facility, contributor base we need to have a `standards document` that encompasses these.

awalter-bnl:
Thanks @danielballan much appreciated

PR #37, opened 2019-01-16
FIX: Declare numpy requirement introduced by 7e549fde.

danielballan:
In #36, a new numpy dependency was introduced. I meant to push this fix-up
commit before merging it minutes ago.

This also adds a requirements.txt and code for parsing it in setup.py, a
pattern copied from our scientific Python cookiecutter.

attn @awalter-bnl

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+jsonschema
+numpy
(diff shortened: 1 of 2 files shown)


awalter-bnl:
Thanks for letting me know @danielballan, I meant to add numpy to requirements but also forgot prior to the last commit.

PR #39, opened 2019-01-22
added run_header schema

gwbischof:


event_model/schemas/run_header.json
```diff
@@ -0,0 +1,25 @@
...
+            "items": {
+                "$ref":"./event_descriptor.json"}},
+        "rescources": { 
+            "type":"array", 
+            "items": {
+                "$ref":"./resource.json"}},
+        "run_stop": {
+            "$ref": "./run_stop.json" }
+    },
+    "required": ["run_start"],
```
tacaswell: uid should be required too?

tacaswell:
I am not sure we want to put this schema (which is an implementation detail of the mongo backing store) and this should live there.

On the other hand, I think that makes the references much harder to manage....

event_model/schemas/run_header.json
gwbischof: I added it

danielballan:
> On the other hand, I think that makes the references much harder to manage....

Looks like jsonschema supports HTTP references, so we can referenced the file in a tagged commit on GH.

@gwbischof I'm going to create a new repo suitcase-mongo for this.

danielballan:
To be continued in http://github.com/NSLS-II/suitcase-mongo

PR #40, opened 2019-01-25
FIX/ENH: Critical fix-ups from ISS work

danielballan:
* Add `compose_event_page`, `compose_datum_page`.
* Fix a critical bug in `Filler`.
* Fix a critical bug in `verify_filled`.

### DIFF
--- a/event_model/test_em.py
+++ b/event_model/test_em.py
@@ -67,22 +67,56 @@ def test_round_trip_pagination():
     res_bundle = run_bundle.compose_resource(
         spec='TIFF', root='/tmp', resource_path='stack.tiff',
         resource_kwargs={})
-    datum_doc = res_bundle.compose_datum(datum_kwargs={'slice': 5})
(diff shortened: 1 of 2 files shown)


event_model/test_em.py
```diff
@@ -313,3 +349,8 @@ def __call__(self, c, d):
         filler('stop', stop_doc)
         assert not filler._closed
     assert event['data']['image'].shape == (5, 5)
+
+    # Test verify_filled.
+    with pytest.raises(event_model.UnfilledData):
+        event_model.verify_filled(event_model.pack_event_page(raw_event))
+    event_model.verify_filled(event_model.pack_event_page(event))
```
mrakitin: Why do we need it twice here? Or should it raise within the `pytest.raises` context manager?

mrakitin: Oh, my bad. I see it's a different event (`raw_event` vs. `event`). Nevermind, looks good!

PR #48, opened 2019-02-27
use object_keys rather than object_name

CJ-Wright:
It seems the name ``object_keys`` is used in bluesky not ``object_name``

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -605,14 +605,14 @@ def compose_event(*, descriptor, event_counter, data, timestamps, seq_num,
 
 def compose_descriptor(*, start, streams, event_counter,
                        name, data_keys, uid=None, time=None,
-                       object_names=None, configuration=None, hints=None,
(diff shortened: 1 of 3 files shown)


CJ-Wright:
@NSLS-II/dama I think this is ready to go (and most likely should go out before the event model tag)

event_model/__init__.py
```diff
@@ -571,8 +571,10 @@ def compose_event_page(*, descriptor, event_counter, data, timestamps, seq_num,
     return doc
 
 
-def compose_event(*, descriptor, event_counter, data, timestamps, seq_num,
+def compose_event(*, descriptor, event_counter, data, timestamps, seq_num=None,
                   filled=None, uid=None, time=None, validate=True):
+    if seq_num is None:
```
danielballan: Oh, thanks for this! I remember hitting this but apparently forgot to go back and fix it.

PR #50, opened 2019-03-05
DOC: Release notes for v1.8.0 and some docstrings

danielballan:


danielballan:
The `compose_*` functions other than `compose_run` are still missing docstrings but they are (1) semi-internal and (2) slightly hard to explain, so I'm declaring victory at this stage for now.

PR #51, opened 2019-03-08
BLD: set python_requires>=3.6

tacaswell:
We use fstrings now

### DIFF
--- a/docs/source/release-history.rst
+++ b/docs/source/release-history.rst
@@ -2,6 +2,13 @@
 Release History
 ***************
 
+v1.8.1 (2019-03-08)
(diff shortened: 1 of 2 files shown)


danielballan:
If I `pip install event-model` with Python 3.5 installed, will it then pick up 1.8.0 because it's still the newest version that doesn't require a specific Python?

mrakitin:
@danielballan, yes, I just tested, python 3.5 picked up event-model 1.8.0.

danielballan:
Then I think that we just need to make sure that this change is released before we drop support for 3.6.

PR #55, opened 2019-03-13
add scan_id to compose_run

CJ-Wright:


awalter-bnl:
> I also think it is, at best, unhelpful to keep a global scan ID that resets to 0 at the beginning of each Python process; users usually want them to be sequential across a longer time than the lifetime of a given Python process.

I strongly agree with this statement, we should not force a reset of `scan_id`

CJ-Wright:
Ok

PR #59, opened 2019-03-18
Fix path_semantics and be robust against empty 'filled'.

gwbischof:


gwbischof:
There are so many commits because the tests fail on my windows machine, so I'm using travis until I can fix the windows errors.

event_model/__init__.py
```diff
@@ -706,7 +708,9 @@ def pack_event_page(*events):
         time_list.append(event['time'])
         uid_list.append(event['uid'])
         seq_num_list.append(event['seq_num'])
-        filled_list.append(event['filled'])
+        filled_list.append(event.get(
+                           'filled',
+                           {key: False for key in event['data'].keys()}))
```
tacaswell: We should only have entries for key that can be filled (that is keys that have an `'external'` field in the descriptor).

gwbischof: Updated

event_model/__init__.py
```diff
@@ -499,6 +499,8 @@ def compose_resource(*, start, spec, root, resource_path, resource_kwargs,
                      path_semantics=os.name, uid=None, validate=True):
     if uid is None:
         uid = str(uuid.uuid4())
+    if path_semantics == 'nt':
```
danielballan: I worry that this might confuse users who look at the function signature (i.e. `compose_resource?` in IPython) and find a default valid that is not a valid value. Have you considered doing this translation from `os.name` to `path_semantics` above the function definition rather than inside it?

```py
default_path_semantics = {...}[os.name]


def compose_resource(..., path_semantics=default_path_semantics, ...)
```

event_model/tests/test_em.py
```diff
@@ -131,8 +146,14 @@ def test_bulk_events_to_event_page():
     desc_bundle_baseline = run_bundle.compose_descriptor(
         data_keys={'motor': {'shape': [], 'dtype': 'number', 'source': '...'}},
         name='baseline')
+
+    if os.name == 'nt':
+        path_root = 'C:\tmp'
```
mrakitin: To avoid it to be evaluated as "tab" (`\t`)
```suggestion
        path_root = 'C:\\tmp'
```

Quick IPython demo:
```py
In [1]: path_root = 'C:\tmp'

In [2]: path_root
Out[2]: 'C:\tmp'

In [3]: print(path_root)
C:	mp

In [4]: path_root = 'C:\\tmp'

In [5]: path_root
Out[5]: 'C:\\tmp'

In [6]: print(path_root)
C:\tmp
```

prjemian:
Isn't there an OS specific python term for the temporary directory? Lacking
that, Windows has an environment variable for TEMPDIR (iirc, need to
confirm exact spelling) so we can relax the assumption on the C: drive.
Otherwise, that is a point of failure, albeit uncommon.

mrakitin:
Thanks for the feedback, @prjemian! I think the motivation here is to test with the Windows path semantics as it would be on a Windows collection machine with real paths, not just temporary placeholders. We had a similar problem in ophyd, https://github.com/NSLS-II/ophyd/pull/703.

A general comment: it would be nice to have a template for the issues/PRs the same way we have it for Bluesky. I hope it will enforce stricter descriptions and motivations sections.

danielballan:
I had the same thought as @prjemian but didn't bother raising it. I'm not sure I understand your point about testing a "real path". I think this would be a reasonable place to use [pytest's `tmp_path` fixture](https://docs.pytest.org/en/latest/tmpdir.html) which is how we handle this in suitcase-utils. It would remove a bunch of `if` blocks and also ensure that the sample files are cleaned up on test exit.

prjemian:
+1 for tmp_path

danielballan:
If you do go with `tmp_path` @gwbischof you will need to specify `pytest >3.9` here: https://github.com/NSLS-II/event-model/blob/873a874b87b75b0a35dc70d9830bc3790f0c7211/requirements-dev.txt#L6

Travis comes with an older version of `pytest` pre-installed, so you need to specifically pin it to prompt `pip` to upgrade.

mrakitin:
@danielballan, it seems the suggested tmp_path fixture will help here. My point was to have test cases, realistically describing different use cases. But I believe it can be obtained via the fixtures by providing different sets of parameters.

danielballan:
OK, sounds good.

event_model/__init__.py
```diff
@@ -861,6 +865,8 @@ def _transpose_list_of_dicts(list_of_dicts):
 
 def _transpose_dict_of_lists(dict_of_lists):
     "Transform dict-of-lists (i.e. DataFrame-like) into list-of-dicts."
+    if dict_of_lists is None:
```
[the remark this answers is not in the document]

event_model/__init__.py

event_model/tests/test_em.py
gwbischof: updated

event_model/tests/test_em.py
```diff
@@ -121,7 +135,7 @@ def test_round_trip_pagination():
     assert actual == expected
 
 
-def test_bulk_events_to_event_page():
+def test_bulk_events_to_event_page(tmp_path):
```
mrakitin: @gwbischof, @danielballan, does it need to be imported somewhere? Or pytest will automatically understand it should come from fixtures?

gwbischof: I think pytest assumes that all arguments for functions that begin with "test" are fixtures.

mrakitin: I see 👍. It is also used the same way in the examples from @danielballan's link in https://github.com/NSLS-II/event-model/pull/59#issuecomment-476302108.

PR #60, opened 2019-03-22
sequence numbers start at 1

CJ-Wright:


CJ-Wright:
@NSLS-II/dama I think this is ready for review, this fixes a bug with document composition, where the sequence number started at 0 rather than 1.

CJ-Wright:
Actually it did come up at XPDD but I didn't realize what was going on. This kept leaving an empty block in grid scans.

CJ-Wright:
@mrakitin I've updated the test

CJ-Wright:
Would it be possible to get a bug fix release after this goes in?

event_model/tests/test_em.py
```diff
@@ -52,9 +52,9 @@ def test_compose_run():
     datum_doc = compose_datum(datum_kwargs={'slice': 5})
     event_doc = compose_event(
         data={'motor': 0, 'image': datum_doc['datum_id']},
-        timestamps={'motor': 0, 'image': 0}, filled={'image': False},
-        seq_num=1)
+        timestamps={'motor': 0, 'image': 0}, filled={'image': False})
     assert 'descriptor' in event_doc
+    assert event_doc['seq_num'] == 1
```
mrakitin: You are making changes in the `compose_descriptor`, shouldn't it test the field exists in `bundle` above?

mrakitin:
Yes, we can tag and build it. As I wasn't closely involved in the paging updates, maybe @tacaswell can also check this PR?

event_model/tests/test_em.py
CJ-Wright: The change is in the `event_counter` which is handed to `compose_event` we don't have access to the `event_counter` until we get to the `event` itself.

mrakitin: I see now. Thanks for the explanation.

tacaswell:
Starting at 0 is more pythonic, but that horse has been out of the stable for 5 years.

If we want to use this to back the RE might as will change the behavior now.

PR #64, opened 2019-03-28
DOC/BLD: add the markdown content type

mrakitin:
## Description

This PR makes the PyPI-published package's description rendered correctly.

## Motivation and Context

By default the rst format is rendered properly at PyPI, but as we have https://github.com/NSLS-II/event-model/blob/master/README.md here, we need to be explicit about the content type.

## How Has This Been Tested?

Not possible to test before actually pushing a new release to PyPI. It was discovered and tested via https://github.com/Nikea/xray-vision/pull/93 --> https://pypi.org/project/xray-vision/.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,7 @@ setup(
     cmdclass=versioneer.get_cmdclass(),
     description="Data model used by the bluesky ecosystem",
     long_description=readme,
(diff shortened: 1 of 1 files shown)


mrakitin:
Thank you :-)

PR #65, opened 2019-04-01
Add RunRouter

danielballan:
Add a tool for routing runs to callbacks based on the content of the
RunStart document and routing event streams to callbacks based on the
content of the RunStart and EventDescriptor documents.

We plan to merge this and release it in an alpha release so that we can
get some experience with it at one or two beamlines. It has been
clearly marked as "experimental"---still open to major changes that may
not be backward-compatible.

This was developed as part of https://github.com/NSLS-II/bluesky/pull/1076
to make callbacks simpler and to handle the possibility of the documents
from multiple runs being interleaved.

This has been tested with just a couple unit tests in this PR. More
should be added before a final release. It has also been exercised with
various callbacks in https://github.com/NSLS-II/bluesky/pull/1076 and
interactively.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -138,6 +138,11 @@ class RunRouter(DocumentRouter):
     It is configured with a list of factory functions that produce callbacks in
     a two-layered scheme, described below.
 
+    .. warning::
+
+       This is experimental. In a future release, it may be changed in a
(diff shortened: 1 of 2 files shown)


danielballan:
Travis passed. Not sure why the webhook didn't fire: https://travis-ci.org/NSLS-II/event-model

PR #70, opened 2019-04-12
Start uid only for composers

CJ-Wright:
Only use start uid in composers so we don't need to cache the entire start

Previously the composers expected then entire start doc just to pull out the
uid, now they expect the uid.


CJ-Wright:
@NSLS-II/dama ready for review

danielballan:
Any thoughts @CJ-Wright? Do you find my reason for keeping it as is compelling? Would adding a code comment to explain the reasoning be a good resolution?

CJ-Wright:
I'll take a comment for now, I'd like to revisit it at some point, but I don't have the bandwidth at the moment.

PR #71, opened 2019-04-19
Added sanitize_np function, which is faster than sanitize_doc

gwbischof:
sanitize_doc was a bottleneck for mongo inserting. sanitize_np fixes this problem. I made it a separate function because sanitize_doc also changes tuples to lists in addition to sanitizing numpy, and this feature is useful for some tests.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -11,6 +11,7 @@ import time as ttime
 import uuid
 import warnings
 from ._version import get_versions
+import numpy as np
 
 __all__ = ['DocumentNames', 'schemas', 'compose_run']
 
@@ -896,6 +897,38 @@ def verify_filled(event_page):
                                    f"`event_model.Filler`.")
 
 
+def sanitize_np(doc):
+    '''Return a copy with any numpy objects converted to built-in Python types.
+    This is a faster version of sanitize_doc which only converts numpy objects.
+
+    This function takes in an event-model document and returns a copy with any
+    numpy objects converted to built-in Python types. It is useful for
+    sanitizing documents prior to sending to any consumer that does not
+    recognize numpy types, such as a MongoDB database or a JSON encoder.
+
+    Parameters
+    ----------
+    doc : dict
+        The event-model document to be sanitized
+
+    Returns
+    -------
+    sanitized_doc : event-model document
+        The event-model document with numpy objects converted to built-in
+        Python types.
+    '''
+    return {key: sanitize_item(value) for key, value in doc.items()}
+
+
+def sanitize_item(val):
+    "Convert any numpy objects into built-in Python types."
+    if isinstance(val, (np.generic, np.ndarray)):
+        if np.isscalar(val):
+            return val.item()
+        return val.tolist()
+    return val
+
+
 def sanitize_doc(doc):
     '''Return a copy with any numpy objects converted to built-in Python types.
 
@@ -936,3 +969,4 @@ class NumpyEncoder(json.JSONEncoder):
                 return obj.item()
             return obj.tolist()
         return json.JSONEncoder.default(self, obj)
+


tacaswell:
This does not look like it will catch:

```python
d = {'a': {'b': np.arange(5))}}

```

danielballan:
Can this be updated to handle a list of arrays? Example:

```py
In [3]: sanitize_np({'a': [numpy.arange(5)]})
Out[3]: {'a': [array([0, 1, 2, 3, 4])]}
```

We encountered that in the wild at some point and that was our reason for switching to the approach currently used in `sanitize_doc`. Would be nice to have something that can handle that but is fast, which a round-trip through JSON is not.

danielballan:
It would be better to duck-type if we can. Resorting to `isinstance` creates fragility because it won't work on things are dict-_like_ or list-_like_. For example, this won't recurse into tuples. See for example https://github.com/NSLS-II/databroker/blob/30e9c707171df5ad14cd58a21382284af7c7db87/databroker/utils.py#L29-L42 where we sniffed out dicts based on the existence of an `.items()` attribute. A good trick for catching lists/tuples/etc. is to use `isinstance(x, collections.abc.Iterable)`.

event_model/__init__.py
```diff
@@ -896,6 +898,51 @@ def verify_filled(event_page):
...
+    Returns
+    -------
+    sanitized_doc : event-model document
+        The event-model document with numpy objects converted to built-in
+        Python types.
+    '''
+
+    def iterate_sanitize(doc):
+        if hasattr(doc, 'items'):
+            for value in doc.values():
```
klauer:
```suggestion
            doc = {key: sanitize_np(value) for key, value in doc.items()}
```

event_model/__init__.py
```diff
@@ -896,6 +898,51 @@ def verify_filled(event_page):
...
+                iterate_sanitize(value)
+        elif isinstance(doc, collections.abc.Iterable) and not isinstance(str):
+            doc = list(doc)  # Change tuples to lists
```
klauer:
```suggestion
            doc = [sanitize_np(value) for value in doc]
```

event_model/__init__.py
```diff
@@ -896,6 +898,51 @@ def verify_filled(event_page):
...
+            for value in doc:
+                iterate_sanitize(value)
+        else:
+            doc = sanitize_item(doc)
+
+    iterate_sanitize(doc)
```
klauer:
```suggestion
    return iterate_sanitize(doc)
```

Is this inner function even necessary? Seems like this outer function (i.e., `sanitize_np`) should be recursive

event_model/__init__.py
gwbischof: Changed a few things just now, so this is fixed

event_model/__init__.py
gwbischof: Changed a few things just now, so this is fixed

event_model/__init__.py
gwbischof: I think my last commit addressed this point. Do you agree?

gwbischof:
I'll add some tests before trying to merge, I'll work on that later.

event_model/__init__.py
```diff
@@ -896,6 +898,45 @@ def verify_filled(event_page):
...
+        Python types.
+    '''
+    if hasattr(doc, 'items'):
+        return {key: sanitize_np(value) for key, value in doc.items()}
+    elif isinstance(doc, collections.abc.Iterable) and not isinstance(doc, str):
+        return [sanitize_np(item) for item in doc]
+    else:
+        return sanitize_item(doc)
+
+    return sanitize_np(doc)
```
klauer: Unreachable line

PR #73, opened 2019-05-02
correct stop doc number of events

CJ-Wright:

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -751,6 +751,7 @@ def compose_event_page(*, descriptor, event_counter, data, timestamps, seq_num,
 def compose_event(*, descriptor, event_counter, data, timestamps, seq_num=None,
                   filled=None, uid=None, time=None, validate=True):
     if seq_num is None:
+        event_counter[descriptor['name']] += 1
         seq_num = event_counter[descriptor['name']]
     if uid is None:
         uid = str(uuid.uuid4())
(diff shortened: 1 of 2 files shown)


CJ-Wright:
@NSLS-II/dama ready for review

CJ-Wright:
This is not in conflict with #60. #60 fixed one problem (the sequence numbers started at 0) but caused another problem (the stop document doesn't reflect the correct number of events). This fixes the second problem. Please see the test here for the exact change: https://github.com/NSLS-II/event-model/pull/73/files#diff-f349740d8f67ede057eaf75fa283f640R60 (This test fails on current master)

event_model/__init__.py
```diff
@@ -751,6 +751,7 @@ def compose_event_page(*, descriptor, event_counter, data, timestamps, seq_num,
 def compose_event(*, descriptor, event_counter, data, timestamps, seq_num=None,
                   filled=None, uid=None, time=None, validate=True):
     if seq_num is None:
+        event_counter[descriptor['name']] += 1
```
danielballan: This will increment the `seq_num` even if the call later errors out below. The original only increments if a document is produced. Can we fix this in `compose_stop` instead?

mrakitin:
Thanks for the description @CJ-Wright. I am seconding @danielballan's suggestion.

CJ-Wright:
I think I have implemented the changes.

CJ-Wright:
Great! Can I get a merge?

CJ-Wright:
Thank you!

PR #74, opened 2019-05-13
CI: Publish to bluesky.github.io.

danielballan:


.travis.yml
```diff
@@ -27,3 +30,4 @@ script:
   # Pubish the documentation to GitHub Pages.
   - pip install doctr
   - doctr deploy --built-docs docs/build/html --deploy-repo NSLS-II/NSLS-II.github.io event-model
+  - doctr deploy --built-docs docs/build/html --deploy-repo bluesky/bluesky.github.io event-model
```
mrakitin: Do we also need `--deploy-branch-name master` as in https://github.com/bluesky/databroker/pull/406/files#diff-354f30a63fb0907d4ad57269548329e3R70?

danielballan: Good catch! I think this is actually OK as is, but I had to check.

I believe that we used to need to specify `--deploy-branch-name` when the doctr first added support for `org.github.io` repos. (I contributed that feature to doctr myself.) But now doctr is smart enough to deploy to `master` if the `--deploy-repo` is a github.io site. See https://github.com/drdoctr/doctr/blob/0f19ff78c8239efcc98d417f36b0a31d9be01ba5/doctr/travis.py#L335-L336 So we should leave this as is and go back and simplify databroker and any others that specify `--deploy-branch-name` because it is now redundant.

PR #79, opened 2019-05-22
build schema_validators dict for bluesky issue #1197

jklynch:
A dict 'schema_validators' is constructed of jsonschema.Validator instances. The intention is that this dict be imported by code wanting to validate JSON.

## Motivation and Context
This change is in support of using jsonschema>3: https://github.com/bluesky/bluesky/issues/1197.

jsonschema.validate() is currently used in bluesky but has been deprecated. The 'new' method is to extend a jsonschema validator class which takes about 4 lines of code. I suggest putting those 4 lines in event-model and caching one validator for each schema.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -635,6 +635,25 @@ for name, filename in SCHEMA_NAMES.items():
     with open(rs_fn('event_model', filename)) as fin:
         schemas[name] = json.load(fin)
 
+
+def _is_array(checker, instance):
+    return (
+        jsonschema.validators.Draft7Validator.TYPE_CHECKER.is_type(instance, 'array') or
+        isinstance(instance, tuple)
+    )
+
+
+_array_type_checker = jsonschema.validators.Draft7Validator.TYPE_CHECKER.redefine('array', _is_array)
+
+
+_Validator = jsonschema.validators.extend(
+    jsonschema.validators.Draft7Validator,
+    type_checker=_array_type_checker)
+
+
+schema_validators = {name: _Validator(schema=schema) for name, schema in schemas.items()}
+
+
 __version__ = get_versions()['version']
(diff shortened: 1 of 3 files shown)


danielballan:
This looks on track so me so I have gone ahead and marked it as ready for review so that CI will start running against it.

Please update all the places that we use `jsonschema.validate` , such as https://github.com/bluesky/event-model/blob/54e925915b29fe77c9207887a73c2562dc2f6578/event_model/__init__.py#L73

to use these validators instead. Please also update the `requirements.txt` to specify the minimum version of `jsonschema` that supports this new usage.

jklynch:
I think I have corrected those embarrassing oversights.

danielballan:
Haha, not at all! Very useful PR for Week 3 on the team!

event_model/__init__.py
```diff
@@ -10,11 +8,13 @@
 import time as ttime
 import uuid
 import warnings
-from ._version import get_versions
 
-__all__ = ['DocumentNames', 'schemas', 'compose_run']
+import jsonschema
+import numpy
```
mrakitin: Why did those imports moved from their original places?

jklynch:
PEP8 compulsion. I was weak.

danielballan:
I don't think we ever really decided whether we would adopt the "builtins, then external libs, then package-local libs" breakdown, but I have heard it argued for and I am happy to adopt it.

danielballan:
Per quick Slack check-in with @mrakitin we are indeed all on the same page that this is a good and reasonable compulsion. Thanks again @jklynch!

mrakitin:
I really like it though, as I expressed it some time ago myself. PyCharm normally does it for you "for free" ("Optimize imports" menu or something like that).

PR #86, opened 2019-06-06
CI: Add 3.7 and nightly builds.

danielballan:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,13 @@
 language: python
-python:
-  - 3.6
+matrix:
+  fast_finish: true
+  include:
+    - python: 3.6
+    - python: 3.7
+      env: PUBLISH_DOCS=1
+    - python: nightly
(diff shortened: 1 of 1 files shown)


mrakitin:
Oh, got it. I think the Ubuntu version is outdated there (v14), that's why it cannot find the Python distribution. I think we saw it with https://github.com/bluesky/bluesky/pull/1203.

.travis.yml
```diff
@@ -1,6 +1,13 @@
 language: python
```
mrakitin:
```suggestion
dist: xenial
language: python
```

danielballan: Thanks, I would have sworn I included that. Not sure how I missed it!

mrakitin: Actually it should be enabled by default. Probably it wasn't picket up due to some weird caching on the TravisCI side.

Issue #89, opened 2019-06-18
Considering a 'fill_item' API


danielballan:
For some work over on intake-bluesky, it has been proposed to add something like `Filler.fill_item(datum_id)` to the `Filler` API. This is reminiscent of `filestore.retrieve(datum_id)` in the original "file store" implementation of filling, which proved problematic. I will summarize some thoughts on this from conversations with @gwbischof and @tacaswell:

## Resource Management

One problem with `filestore.retrieve(datum_id)` is that the caches involved were global to the `filestore` module. The `Filler` is a class (and a context manager to boot). It would not have this problem.

## Keeping Context Next to Data

Another problem with `filestore.retrieve(datum_id)` is that is returns a bare array, rather than a dict that keeps labels next to their values. This is usability issue---it's not unsolvable by any means, but it puts some burden on the caller to keep track of what is what.

As noted by @gwbischof, in the context of a dask-based filler, dask would track this state for us. Filling is generally such a low-level thing that we probably shouldn't worry about this too much.

## Laying a Path For Batched Filling

Because the handler API only accepts one Datum at a time, filling blocks of data unavoidably involved a hot loop over a function call `handler(**datum_kwargs)`. At some point we will need to extend the handler API to enable more efficient filling, along the lines of `handler.fill_a_bunch_of_stuff_in_one_call(???)`. There is not an obvious way to do that yet, but we'll need to figure it out eventually.

In @tacaswell's judgement, adding new API with a signature like `fill_item(datum_id)` is likely to make such future changes harder. We should focus on filling pages/chunks as a unit and then accessing the specific data we need by indexing into the page/chunk. I find that argument convincing.

### Add `fill_event` and `fill_event_page`.

This makes it possible to use a `Filler` to specifically fill certain columns in certain events, without breaking its utility as a normal `DocumentRouter`-based callback.

```python
class Filler(...):
    def fill_event(self, doc, include=None, exclude=None):
        # Move the guts of `Filler.event` into here.

    def fill_event_page(self, doc, include=None, exclude=None):
        # Loop over fill_event. In the future, when the handler API is extended,
        # we may be able to actually fill event pages efficiently without unpacking.
```

We can then rewrite `event` and `event_page` to use these methods.

Issue #88, opened 2019-06-17
Add option to *not* fill in place, and later make it default.


danielballan:
In early days of databroker, we made the design to decision that filling should mutate Event documents in place, replacing datum ID with the referenced data. We now view this as a mistake. Consider the situation where we have three consumers subscribed to the RunEngine:

1. Writes documents to MongoDB
2. Fills documents (mutates in place)
3. Writes image data to TIFF files

Note that (1) assumes unfilled documents and (3) assumes filled documents. If these are subscribed in any other order, these expectations will not be met, and things will fail badly. It would be safer for the filling process to return a _copy_ that can be explicitly passed on to downstream consumers that expect filled data.

Filling in place does have an advantage: by avoiding the copying, it can be faster. But it would be safer to make this an _opt in_ behavior.

I propose adding an `inplace` argument to `Filler.__init__`. In the first release, the default value will be `None`, a signal that the user has not told us what they want. Then, if it is found to be `None`, a *warning* will advise the user to explicit specify `True` or `False`, but it will behave in a behavior-compatible way. (That is, it will still fill in place.) In a later release, we will change the default from `None` to `False`. This will break any existing code that has not been updated---hence warning for a release or two.

PR #90, opened 2019-06-18
added fill_event, and fill_event_page methods

gwbischof:


gwbischof:
Fixes https://github.com/bluesky/event-model/issues/89

gwbischof:
Fixes https://github.com/bluesky/event-model/issues/88

event_model/__init__.py
```diff
@@ -390,9 +390,19 @@ class Filler(DocumentRouter):
...
+                 descriptor_cache=None, inplace=None,
                  retry_intervals=(0.001, 0.002, 0.004, 0.008, 0.016, 0.032,
                                   0.064, 0.128, 0.256, 0.512, 1.024)):
+
+        if inplace is None:
+            self._inplace = True
+            warnings.warn(
+                "'inplace' argument not specified. It is recommended to "
+                "specify True or False. In future releases, 'inplace' "
+                " will default to False.")
```
danielballan:
```suggestion
                "will default to False.")
```

event_model/__init__.py
```diff
@@ -390,9 +390,19 @@ class Filler(DocumentRouter):
     def __init__(self, handler_registry, *,
                  include=None, exclude=None, root_map=None,
                  handler_cache=None, resource_cache=None, datum_cache=None,
-                 descriptor_cache=None,
+                 descriptor_cache=None, inplace=None,
                  retry_intervals=(0.001, 0.002, 0.004, 0.008, 0.016, 0.032,
                                   0.064, 0.128, 0.256, 0.512, 1.024)):
+
+        if inplace is None:
+            self._inplace = True
```
danielballan: I like that this is internal (i.e. has an underscore). Should we have a public, read-only property so that the user can check whether a given `FIller` instance is in-place or not?

danielballan:
Another useful test to add:

```python
with pytest.warns(UserWarning):
    # Make a Filler with inplace=None
```

event_model/__init__.py

event_model/__init__.py
gwbischof: added

event_model/__init__.py
```diff
@@ -465,19 +480,37 @@ def event_page(self, doc):
...
+            doc['data'] = filled_doc['data']
+            doc['filled'] = filled_doc['filled']
+            return doc
+        else:
+            return filled_doc
 
     def event(self, doc):
+        filled_doc = self.fill_event(doc, include=self.include,
+                                     exclude=self.exclude)
+        if self._inplace:
```
tacaswell: Would it make more sense to do this logic it `fill_event` (and `fill_event_page`) so that if the user wants inplace we can skip the deepcopy all together an so that `event` and `fill_event` follow the same `inplace` rules?

event_model/tests/test_em.py
```diff
@@ -455,6 +456,30 @@ def __call__(self, c, d):
...
+        assert filled_event is event
+
+    with event_model.Filler(reg, inplace=False) as filler:
+        filler('start', run_bundle.start_doc)
+        filler('descriptor', desc_bundle.descriptor_doc)
+        filler('descriptor', desc_bundle_baseline.descriptor_doc)
+        filler('resource', res_bundle.resource_doc)
+        filler('datum', datum_doc)
+        event = copy.deepcopy(raw_event)
+        filled_event = filler.event(event)
```
tacaswell: Can you use the same `filler('event', event)` pattern as the rest of the calls above use?

event_model/__init__.py
```diff
@@ -390,9 +391,19 @@ class Filler(DocumentRouter):
     def __init__(self, handler_registry, *,
                  include=None, exclude=None, root_map=None,
                  handler_cache=None, resource_cache=None, datum_cache=None,
-                 descriptor_cache=None,
+                 descriptor_cache=None, inplace=None,
                  retry_intervals=(0.001, 0.002, 0.004, 0.008, 0.016, 0.032,
                                   0.064, 0.128, 0.256, 0.512, 1.024)):
+
+        if inplace is None:
+            self._inplace = True
```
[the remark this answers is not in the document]
danielballan: Good question. The current behavior is equivalent to `inplace=True`, so changing it to `False` will be an API break. That's why we plan to issue warnings for at least one release before breaking user code.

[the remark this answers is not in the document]
mrakitin: OK, just wanted to be sure it's consistent with the message a user will see. So the eventual goal is to have the default to be `False`, right?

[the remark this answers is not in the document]
danielballan: Yes. Any suggestions to make the warning message more clear?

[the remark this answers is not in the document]
mrakitin: No, it all looks good. Thanks for the confirmation.

Issue #49, opened 2019-03-05
Add rechunk_event_page, rechunk_datum_page.


danielballan:
[Edited per first comment below]

```py
def rechunk_event_pages(event_pages, chunk_size):
    "Yield chunks with up to chunk_size Events."
    ...
```

tacaswell:
May be more useful to take a generator of event_pages so that we can use this to make _bigger_ event pages as well as smaller one.s

awalter-bnl:
This sounds like a good idea.

danielballan:
We might also want a coroutine version of this. This is, in effect, the `Paginator` component that we had planned to build once https://github.com/NSLS-II/bluesky/pull/1125 finally lands.

PR #92, opened 2019-06-20
added rechunk_event_pages

gwbischof:
Fixes https://github.com/bluesky/event-model/issues/49


event_model/__init__.py
```diff
@@ -1045,6 +1045,73 @@ def unpack_datum_page(datum_page):
...
+                **{'data': page['data'][key][start:stop]
+                   for key in page['data'].keys()}.values())},
+                **{'timestamps': page['timestamps'][key][start: stop]
+                   for key in page['data'].keys()}.values())},
+                **{'filled': page['filled'][key][start:stop]
+                   for key in page['data'].keys()}.values())}}
+
+
+def merge_event_pages(pages):
+    if len(pages) == 1:
```
danielballan: It might not have a length. (Generators would raise a `TypeError` here.)

event_model/__init__.py
```diff
@@ -1045,6 +1045,73 @@ def unpack_datum_page(datum_page):
...
+        return pages[0]
+
+    array_keys = ['seq_num', 'time', 'uid']
+
+    return {**{'descriptor': pages[0]['descriptor']},
+            **{key: list(itertools.chain.from_iterable(
```
danielballan: This use of `**` is elegant. Nice!

event_model/__init__.py
```diff
@@ -1045,6 +1045,73 @@ def unpack_datum_page(datum_page):
...
+                    [page[key] for page in pages]) for key in array_keys])},
+            **{'data': {key: list(itertools.chain.from_iterable(
```
danielballan: ...But here and below, I don't think `**` is buying you anything. `{..., **{'data': ...}}` can also be written as `{..., 'data': ...}`, right?

event_model/__init__.py
```diff
@@ -1045,6 +1045,73 @@ def unpack_datum_page(datum_page):
...
+        for chunk in new_chunks:
+            remainder -= len(chunk['uid'])
+            chunk_list.append(chunk)
+            if remainder == 0:
+                yield merge_event_pages(chunk_list)
+                remainder = chunk_size
+    if chunk_list:
+        yield merge_event_pages(chunk_list)
+
+    def page_chunks(page, chunk_size, remainder):
```
danielballan: Any reason not to make this a top-level private function `_page_chunks`? That would enable us to test it separately. In-lining in a good way to do closures, which I don't think we need here, or to ensure that something definitely isn't exposed to users, which doesn't seem like an issue here either.

event_model/__init__.py
```diff
@@ -1045,6 +1045,73 @@ def unpack_datum_page(datum_page):
...
+def rechunk_event_pages(event_pages, chunk_size):
+    remainder = chunk_size
+    chunk_list = []
+
+    for page in event_pages:
+        new_chunks = page_chunks(page, chunk_size, remainder)
+        for chunk in new_chunks:
+            remainder -= len(chunk['uid'])
+            chunk_list.append(chunk)
+            if remainder == 0:
```
danielballan: Should this be `<=`? (What happens if `page_size` isn't divisible by `len(chunk['uid'])`. A good test case to add.

event_model/__init__.py
```diff
@@ -1045,6 +1045,73 @@ def unpack_datum_page(datum_page):
...
+                    for key in page['data'].keys()}},
+            **{'timestamps': {key: list(itertools.chain.from_iterable(
+                    [page['timestamps'][key] for page in pages]))
+                    for key in page['data'].keys()}},
+            **{'filled': {key: list(itertools.chain.from_iterable(
+                    [page['filled'][key] for page in pages]))
+                    for key in page['data'].keys()}}}
+
+
+def merge_datum_pages(pages):
```
danielballan: Will there be a `rechunk_datum_pages` also?

event_model/__init__.py
gwbischof: I think we want ==, but I'm not totally sure. I don't think remainder will ever be negative, even if page size isn't evenly divisible by chunk size. That's what I intended at least. Good idea, ill write a test case for this.

event_model/__init__.py
gwbischof: yea, ill add `rechunk_datum_pages`

event_model/__init__.py
gwbischof: The reason I planned to nest it because I thought, people shouldn't be using it directly. `page_chunks` doesn't give even sized chunks, the first chunk will be of size remainder. If someone wants even sized chunks of a page they could use `rechunk_event_pages`. Are there some uses for a chunking function that has a remainder argument?

event_model/__init__.py
gwbischof: thanks!

event_model/__init__.py
gwbischof: I'm not sure, I'll try that out. Thanks for all of the suggestions.

event_model/__init__.py

event_model/__init__.py

event_model/__init__.py

event_model/tests/test_em.py
```diff
@@ -481,6 +481,55 @@ def __call__(self, c, d):
...
+            yield {'descriptor': 'DESCRIPTOR',
+                   **{key: list(range(page_size)) for key in array_keys},
+                   'data': {key: list(range(page_size)) for key in data_keys},
+                   'timestamps': {key: list(range(page_size)) for key in data_keys},
+                   'filled': {key: list(range(page_size)) for key in data_keys}}
+
+    # Get a list of event pages of size 13.
+    event_pages = list(event_page_gen(13, 31))
+    # Change the size of the event_pages to size 7.
+    event_pages_7 = event_model.rechunk_event_pages(event_pages, 7)
```
danielballan: I think it would be good to check here that the sizes of the pages in `event_pages_7` are `57 * [4] + [4]` as expected. From interactive testing I believe this is working correctly, but for the sake of future-proofing against future changes we should verify that here.

event_model/tests/test_em.py
```diff
@@ -481,6 +481,55 @@ def __call__(self, c, d):
...
+    # Change the size back to 13.
+    event_pages_13 = event_model.rechunk_event_pages(event_pages_7, 13)
```
danielballan: ...and also here.

event_model/tests/test_em.py
```diff
@@ -481,6 +481,55 @@ def __call__(self, c, d):
...
+        for i in range(num_pages):
+            yield {'resource': 'RESOURCE',
+                   **{key: list(range(page_size)) for key in array_keys},
+                   'datum_kwargs': {key: list(range(page_size))
+                                    for key in data_keys}}
+
+    # Get a list of datum pages of size 13.
+    datum_pages = list(datum_page_gen(13, 31))
+    # Change the size of the datum_pages to size 7.
+    datum_pages_7 = event_model.rechunk_datum_pages(datum_pages, 7)
```
danielballan: Same as above.

event_model/tests/test_em.py
gwbischof: updated

event_model/tests/test_em.py
gwbischof: updated

event_model/tests/test_em.py
gwbischof: I think it would be redundant to add it here, because we are already checking that it matches the original exactly.

danielballan: Oh good point.

Issue #78, opened 2019-05-15
documentation update: descriptor and resource details


gwbischof:
I think the documentation for descriptors and resources should be improved:
http://nsls-ii.github.io/event-model/data-model.html#resource-document
http://nsls-ii.github.io/event-model/data-model.html#event-descriptor

It isn't clear how to understand the meaning of the different parts of these documents.

gwbischof:
https://nsls-ii.github.io/bluesky/event_descriptors.html#

Here is some documentation already created for descriptors

danielballan:
I agree that the documentation should be improved. Action item here should be to port the narrative text from https://blueskyproject.io/bluesky/documents.html and https://nsls-ii.github.io/architecture-overview.html and https://nsls-ii.github.io/bluesky/event_descriptors.html into this documentation and then delete those so that we can maintain it all in one place.

jklynch:
I'm looking at https://blueskyproject.io/bluesky/documents.html. Should this file go in to the event-model documentation as-is, or should it be combined with an existing file? Also it references the callbacks page in the bluesky documentation. Should it continue to reference that?

danielballan:
My sense of it at this point is:

- https://blueskyproject.io/bluesky/documents.html should be combined into https://blueskyproject.io/event-model/data-model.html. The "Overview of a Run" section can just be dropped, I think, but the bulleted lists that describe the contents of each document type are worth merging into the cooresponding sections in data-model.html. Any links to sections of the bluesky documentation should probably be dropped.
- The descriptors page https://nsls-ii.github.io/bluesky/event_descriptors.html should be moved wholesale to event-model and adapted if needed to remove bluesky links.
- This paragraph from the architecture-overview.html should be merged in as well:

> The scope of a “run” is left to the beamline or individual scientist. It is quite generic: a set of Documents generated by following a given sequence of instructions. This might encompass a single short measurement (say, “a count”) or a multi-step procedure such as a raster scan. A run should represent a set of individual measurements that will be collected or processed together.

PR #94, opened 2019-06-29
Doc issue 78

jklynch:
Move event-model documentation from bluesky to event-model.

## Description
Merge https://blueskyproject.io/bluesky/documents.html into https://blueskyproject.io/event-model/data-model.html.

Move https://nsls-ii.github.io/bluesky/event_descriptors.html into the event-model project.

## Motivation and Context
This PR addresses issue #78.


jklynch:
I left some text between "begin remove" ... "end remove" markers.

docs/source/data-model.rst
```diff
@@ -1,18 +1,39 @@
-**********
-Data Model
-**********
+********************************
+Data Model: Documents and Events
```
danielballan: I think of Events as one of the kinds of Documents.

docs/source/data-model.rst
```diff
@@ -1,18 +1,39 @@
...
 
+*end remove*
+
+A primary design goal of bluesky is to enable better research by recording
+rich metadata alongside measured data for use in later analysis. Documents are
+how we do this.
+
+A *document* is our term for a Python dictionary with a schema --- that is,
+organized in a
+`formally specified <https://github.com/NSLS-II/event-model>`_ way --- created
```
danielballan: This link is a cross-reference from bluesky to event-model. Now that we are in event-model, I think we can just remove this link.

docs/source/data-model.rst
```diff
@@ -237,10 +354,16 @@ can always be losslessly transformed into an Event and vice versa.
...
+A 'stop' document marks the end of the run. It contains metadata that is not
+known until the run completes.
+
+The most commonly useful fields here are 'time' and 'exit_status'.
+
 Minimal nontrivial valid example:
 
 .. code-block:: python
 
+   # 'stop' document
```
danielballan: Do you think we should add a comment like this to _all_ the examples? Most of them do not have one. I would rather apply it to none or all.

docs/source/data-model.rst
```diff
@@ -1,18 +1,39 @@
-**********
-Data Model
-**********
+********************************
+Data Model: Documents and Events
+********************************
+
+*begin remove*
```
danielballan: I still like this original first paragraph, doesn't seem exactly redundant to what follows. Maybe combine it with the next paragraph?

docs/source/data-model.rst
```diff
@@ -1,18 +1,39 @@
...
+rich metadata alongside measured data for use in later analysis. Documents are
+how we do this.
+
+A *document* is our term for a Python dictionary with a schema --- that is,
+organized in a
+`formally specified <https://github.com/NSLS-II/event-model>`_ way --- created
+by the RunEngine during plan execution.  All of the metadata and data generated
+by executing the plan is organized into documents.
+
+A :doc:"later section <callbacks>" describes how outside functions can
```
danielballan: This is one of the bluesky internal references of the kind I think we should just remove. Maybe this whole paragraph can go?

docs/source/event-descriptors.rst
```diff
@@ -0,0 +1,210 @@
+Event Descriptors
+=================
+
+In the section on :doc:`documents`, we gave an overview of the four kinds of
```
danielballan:
```suggestion
In the section on :doc:`data-model`, we gave an overview of the types of
```

docs/source/event-descriptors.rst
```diff
@@ -0,0 +1,210 @@
...
+document. We presented an example Run Start, Event, and Run Stop, but we
```
danielballan: This second sentence is no longer accurate. We do present an Event Descriptor example. We just down break it down and discuss it in detail.

docs/source/event-descriptors.rst
```diff
@@ -0,0 +1,210 @@
...
+deferred detailed discussion of the Event Descriptor.
+
+Recall our example 'event' document.
+
+.. code-block:: python
+
+    # 'event' document (same as above, shown again for reference)
```
danielballan: "above" is out of date --- this is own its own page

docs/source/data-model.rst
```diff
@@ -1,18 +1,39 @@
...
+a sense of the structure and familiarity with useful components.
+
 Overview
 ========
 
 The data model is composed of six types of Documents, which in Python are
 represented as dictionaries but could be represented as nested mappings (e.g.
 JSON) in any language. Each document class has a defined, but flexible, schema.
 
+*begin remove*
```
danielballan: I'm surprised you marked this for removal. I figure it's good to have a very short summary of the meaning of each document somewhere. As someone coming to this fresh, do you think this is so succinct that it doesn't add much value, that we might as well go straight to the more detailed description?

I would be happy with any of the following:
- Drop bluesky's version of this overview and keep event-model's.
- Merge some of event-model's verbiage (e.g. "the who / why / what") into bluesky's.

It depends on how much detail we want to present this early.

docs/source/data-model.rst
```diff
@@ -97,7 +165,46 @@ Typical example:
...
+    from bluesky.plans import scan
+    from ophyd.sim import det, motor  # simulated detector, motor
+
+    # Scan 'motor' from -3 to 3 in 10 steps, taking readings from 'det'.
+    RE(scan([det], motor, -3, 3, 16), purpose='calibration',
+       sample='kryptonite')
+
+generates a typical 'run start' document like this:
+
+.. code-block:: python
```
danielballan: This is older and less accurate that the version in event-model. We should keep the version in event-model. We could however reincorporate the example bluesky code if that seems helpful.

docs/source/data-model.rst
jklynch: I have had some confusion with the event/Event terminology. In my mind there are event-based execution model "events" and then there are "Event Documents". I try not to think of the documents as "events" but as "Events". I think you are right, this heading should go back to just "Data Model".

docs/source/data-model.rst
jklynch: removed

docs/source/data-model.rst
prjemian: So, let's be clearer.  Instead of starting with the next paragraph, start with something more direct.

```
In data acquisition, an *event* is a set of readings coordinated at a specific time.  A primary design goal of bluesky is to enable better research by recording rich metadata alongside the event data for use in later analysis.  Bluesky encapsulates all information from each event in a *Document*.  Other types of Document describe the events associated with an experimental run.
```

docs/source/data-model.rst
prjemian: AND, leave the title of this section as "Data Model: Documents and Events" since it creates the necessary association with the data acquisition term *event* and the bluesky document type *event*.

docs/source/data-model.rst
jklynch: I added them to all the examples.

docs/source/data-model.rst
jklynch: I spent some time trying to combine the original first paragraph and the text from documents.rst. My problem is that the original is talking about little-e "events" and it seems confusing to talk about big-e "Events" at the same time.

I will try again.

jklynch: Here is a new proposal: I added the original paragraph to the second new paragraph. I still think using the term "event-based" is confusing so I propose using "document-based" instead. I felt like the original text fits here since the following paragraph talks about subscribing to a stream of documents.

```
A primary design goal of bluesky is to enable better research by recording
rich metadata alongside measured data for use in later analysis. Documents are
how we do this. 

A *document* is our term for a Python dictionary with a schema created
by the RunEngine during plan execution.  All of the metadata and data generated
by executing the plan is organized into documents. Bluesky's document-based
data model supports complex, asynchronous data collection and enables
sophisticated live, prompt, streaming, and *post-facto* data analysis.

A :doc:"later section <callbacks>" describes how outside functions can
"subscribe" to a stream of these documents, visualizing, processing, or saving
them. This section provides an outline of documents themselves, aiming to give
a sense of the structure and familiarity with useful components.

```

jklynch: working on the third paragraph below....

docs/source/data-model.rst
jklynch: I liked the mention of subscribing to a document stream. It convinces me that this is very useful information. I propose to link to the Bluesky documentation:
```
The `bluesky documentation <https://blueskyproject.io/bluesky/>`_ describes
how outside functions can "subscribe" to a stream of these documents,
visualizing, processing, or saving them. This section provides an outline of
documents themselves, aiming to give a sense of the structure and familiarity
with useful components.
```

docs/source/event-descriptors.rst
jklynch: check

docs/source/event-descriptors.rst
jklynch: check

docs/source/event-descriptors.rst
jklynch: check

docs/source/data-model.rst
jklynch: check

docs/source/data-model.rst
jklynch: I liked the similar summary of each document from documents.rst because there was a lot of detail, but I see the value of a shorter summary for each document. I will drop the bluesky version.

jklynch:
In the "Overview" section at the top of data-model.rst there is no summary for resource or datum documents. @danielballan if you would like to write those I will put them in.

docs/source/data-model.rst
danielballan: That makes a lot of sense.

docs/source/data-model.rst
danielballan: Hmm, I agree this is tricky. This line once made sense

> created by the RunEngine during plan execution

but is now too limiting: we can "ingest" data files from non-bluesky beamlines or simulations to create documents now. I wouldn't want to give the wrong impression that the RunEngine is the only way to mint documents.

danielballan:
They are there; they are just below the diagram since they aren't necessary for understanding the diagram.

image

Would it better to move them up?

jklynch:
The first sentence in the Overview section is "The data model is composed of six types of Documents...." so I feel let down because the list has only four. I would like to move the Resource and Datum documents up.

danielballan:
Sounds good.

tacaswell:
"Reference" is not much better than "Pointer" in terms of clarity. It's not really a `uri` because it is structured rather than a string.  

It is "A bundle of data you can trade with another system to get the data you really want" (which is how I understand more-than-one-reference-deep pointers in c++ :stuck_out_tongue_winking_eye: ).

danielballan:
I'm with @prjemian that "pointer" has the potential to confuse sophisticated programmers who, in first encountering our work, provisionally map "pointer" to "memory reference" and need more context to be sure that's not what we mean. I have some empirical evidence for this: I avoided using that term at first, then tried using it for awhile, and found that it confused people. Maybe "record that points to" in place of "pointer to"?

jklynch:
I suggest opening a new issue on the topic of "pointer" vs "reference" vs something else since this PR is focused on moving documentation from the bluesky repo to the event-model repo.

docs/source/data-model.rst
```diff
@@ -20,31 +32,39 @@ JSON) in any language. Each document class has a defined, but flexible, schema.
 * Event Descriptor --- Metadata about a series of Events. Envision
   richly-detail column headings in a table, encompassing physical units,
   hardware configuration information, etc.
+* Resource --- A pointer to an external file (or resource in general).
+* Datum --- A pointer to a specific slice of data within a Resource.
 * Run Stop Document ---  Everything that we can only know at the very end, such
   as the time it ended and the exit status (succeeded, aborted, failed due to
   error).
 
+Every document has a ``time`` (its creation time) and a separate ``uid`` to
```
danielballan: This is not strictly true anymore, now that Resource and Datum are proper documents. They do not have a ``time``, and datum calls its unique id `datum_id`, an unfortunate historical choice that we are more or less stuck with. Maybe, to keep things simple, remove the mention of ``time`` here and just say, "Every document contains a unique identifier."

docs/source/data-model.rst
```diff
@@ -20,31 +32,39 @@ JSON) in any language. Each document class has a defined, but flexible, schema.
...
 rows are contained in one structure for efficient transport and vectorized
 computation. The representations contain equivalent information: an EventPage
 can always be losslessly transformed into an Event and vice versa.
 
-The scope of a "run" is left to the user. It is quite generic: a set of
-Documents generated by following a given sequence of instructions. This might
-encompass a single short measurement (say, "a scan") or a multi-step procedure.
-It can be whatever scope is convenient for accessing a unit of data for later
-analysis.
+The scope of a "run" is left to the beamline or individual scientist. It is
```
danielballan: We have users in sciences outside the synchrotron world, and we go out of our way to avoid creating the impression that bluesky is for synchrotrons only.

```suggestion
The scope of a "run" is left to the research group or individual scientist. It is
```

docs/source/data-model.rst
```diff
@@ -2,9 +2,21 @@
...
 **********
 
-Bluesky's event-based data model supports complex, asynchronous data collection
-and enables sophisticated live, prompt, streaming, and *post-facto* data
-analysis.
+A primary design goal of bluesky is to enable better research by recording
+rich metadata alongside measured data for use in later analysis. Documents are
+how we do this.
+
+A *document* is our term for a Python dictionary with a schema created
```
danielballan: This language should be updated since the RunEngine is now not the only way to create documents. Something like:

> A *document* is our term for a Python dictionary with a schema. The bluesky RunEngine emits documents during plan execution.

We _might_ want to elaborate:

> There are tools for composing documents directly, which can be useful for ingesting data that was not acquired with bluesky into this data model.

But that might be too much detail for this section.

docs/source/data-model.rst
prjemian: This is important language.  Tripped over this before, myself.

docs/source/data-model.rst
prjemian: > ... is left to the instrument team.

docs/source/data-model.rst
```diff
@@ -20,31 +32,39 @@ JSON) in any language. Each document class has a defined, but flexible, schema.
...
   richly-detail column headings in a table, encompassing physical units,
   hardware configuration information, etc.
+* Resource --- A pointer to an external file (or resource in general).
+* Datum --- A pointer to a specific slice of data within a Resource.
 * Run Stop Document ---  Everything that we can only know at the very end, such
   as the time it ended and the exit status (succeeded, aborted, failed due to
   error).
 
+Every document has a ``time`` (its creation time) and a separate ``uid`` to
+identify it. The Event documents also have a ``descriptor`` field linking them
```
prjemian: When this was all new to me, the two terms "Event Descriptor" and "`descriptor`" as a document schema did not jump out as being the same thing.  *Run Start*/`start` and *Run Stop*/`stop` similar but somehow easier to grasp.  Gradually, this struck home.

Is this the time to make this crystal clear for the neophytes?  The new terminology (with familiar names) in bluesky is daunting to new users.

docs/source/data-model.rst
prjemian: *instrument team* has been easier to use and includes anyone who has in interest in this subject - also not specific to "beam line", could be a lab

docs/source/data-model.rst
```diff
@@ -62,6 +82,10 @@ For each type, we will show:
 Run Start Document
 ------------------
 
+Again, a 'run start' document marks the beginning of the run. It comprises
```
prjemian: Here we go again: `'run start'`.  There is no such `run start` document in my stream.  How is that different from a `'start'` document?  Ah, yes, they are the same.

docs/source/data-model.rst
danielballan: Perfect, those are the words I was looking for.

docs/source/data-model.rst
```diff
@@ -71,13 +95,15 @@ Minimal nontrivial valid example:
...
+   # 'run start' document
    {'time': 1550069716.5092213,  # UNIX epoch (seconds since 1 Jan 1970)
     'uid': '10bf6945-4afd-43ca-af36-6ad8f3540bcd'}  # globally unique ID
 
-Typical example:
+A typical example
 
 .. code-block:: python
 
+   # 'run start' document
```
prjemian: ---: # 'run start' document
+++: # run 'start' document

docs/source/data-model.rst
```diff
@@ -106,10 +137,17 @@ Formal schema:
...
+in the Event documents. It provides useful information about each key in the
+data and about the configuration of the hardware. The layout of a descriptor
+is detailed and takes some time to cover, so we defer it to a
+:doc:`later section <event-descriptors>`.
+
 Minimal nontrivial valid example:
 
 .. code-block:: python
 
+   # 'event descriptor' document
```
prjemian: ---: # 'event descriptor' document
+++: # event 'descriptor' document

docs/source/data-model.rst
```diff
@@ -237,10 +283,16 @@ can always be losslessly transformed into an Event and vice versa.
 Run Stop Document
 -----------------
 
+A 'run stop' document marks the end of the run. It contains metadata that is not
```
prjemian: ---: 'run stop'
+++: run 'stop'

event_model/schemas/event_descriptor.json
```diff
@@ -106,7 +106,7 @@
             },
             "run_start": {
                 "type": "string",
-                "description": "Globally unique ID to the run_start document this descriptor is associtaed with."
+                "description": "Globally unique ID to the run_start document this descriptor is associated with."
```
prjemian: truly awkward wording, How about

Globally unique ID of this run's 'start' document.

docs/source/data-model.rst
prjemian: run 'start'

docs/source/data-model.rst
jklynch: check

docs/source/data-model.rst
jklynch: check

docs/source/data-model.rst
jklynch: I am inclined to leave out the elaboration.

docs/source/data-model.rst
jklynch: It would be valuable to pick one term from each pair and stick with it throughout this documentation. My preferences are "Event Descriptor", "Start", and "Stop" for what it's worth. But I think a new PR should be opened for this change.

event_model/schemas/event_descriptor.json
jklynch: check

docs/source/data-model.rst
jklynch: new PR?

docs/source/data-model.rst
jklynch: new PR?

docs/source/data-model.rst
jklynch: new PR?

docs/source/data-model.rst
jklynch: new PR?

docs/source/data-model.rst
prjemian: I agree that is too much for this context.

docs/source/data-model.rst
prjemian: I agree.  A new PR *after* this is all merged.

docs/source/data-model.rst
prjemian: op cit

event_model/schemas/event_descriptor.json
prjemian: I don't see this text in the revision yet: *Globally unique ID of this run's 'start' document.*

docs/source/data-model.rst
danielballan: Great. I lean that way as well, though I included it for consideration.

jklynch:
Sorry @prjemian I don't know why that change wasn't showing up.

PR #95, opened 2019-07-05
FIX/TST: Fix inplace behavior and test it better.

danielballan:
We found that ``fill_event_page`` was not mutating the user input even
when it was told to do so.


danielballan:
@gwbischof  knocked that out quickly to enable some work on intake-bluesky. I will tag a _beta_ release to make it available. If there is a problem in here, we can walk it back before the final release.

PR #96, opened 2019-07-05
CI: Only publish docs on one of the builds.

danielballan:


danielballan:
Hm, perhaps not. I am not sure. I have pushed a commit to check....

.travis.yml
```diff
@@ -35,7 +35,10 @@ script:
   - flake8 --max-line-length=115  # Enforce code style (but relax line length limit a bit).
   - set -e
   - make -C docs html  # Build the documentation.
-  # Pubish the documentation to GitHub Pages.
-  - pip install doctr
-  - doctr deploy --built-docs docs/build/html --deploy-repo NSLS-II/NSLS-II.github.io event-model
-  - doctr deploy --built-docs docs/build/html --deploy-repo bluesky/bluesky.github.io event-model
+  - |
+    if [ $PUBLISH_DOCS ] then
```
mrakitin: Thanks. I think this is missing:
```suggestion
    if [ $PUBLISH_DOCS ]; then
```

mrakitin: Sorry I wasn't precise about the location of the semicolons. It's still should be a part of `if ... then`.

danielballan:
Oops, that makes sense. You can see my bash-fu could use some work. :-D Fixed and force-pushed.

PR #97, opened 2019-07-05
Added validate_order method

gwbischof:
Method to check that documents are generated in an acceptable order.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1354,3 +1354,43 @@ class NumpyEncoder(json.JSONEncoder):
                 return obj.item()
             return obj.tolist()
         return json.JSONEncoder.default(self, obj)
+
+def validate_order(run_gen):
+
+    run_dict = {'start': None,
+                'stop': None,
+                'event': defaultdict(list)}
+
+    last_index = 1
+    for i, namedoc in enumerate(run):
+        name, doc  = namedoc
+        last_index = i
+        if name in {'start', 'stop'}: run_dict[name] = (i, doc)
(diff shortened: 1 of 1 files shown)


gwbischof:
Fixes https://github.com/bluesky/event-model/issues/98

event_model/__init__.py
```diff
@@ -1354,3 +1354,103 @@ def default(self, obj):
...
         return json.JSONEncoder.default(self, obj)
+
+
+class OrderError(ValueError):
+    ...
+
+
+def validate_order(run_iterable):
+    """
+    Validates the order of a Bluesky Run.
```
danielballan: Can you copy over the content from #98 here to explain what constraints this is enforcing?

event_model/__init__.py
```diff
@@ -1354,3 +1354,103 @@ def default(self, obj):
...
+            datum_cache[doc['datum_id']] = doc
+            datum_check(doc)
+        if name == 'datum_page':
+            for datum in unpack_datum_page(doc):
+                datum_cache[datum['datum_id']] = datum
+                datum_check(datum)
+        if name == 'event':
+            event_check(doc)
+        if name == 'event_page':
+            for event in unpack_event_page(doc):
```
danielballan: This does not address:

> Event[Page]s across streams are in time order up to the time resolution of a Page. That is, if we denote event_page['time'][0] as a_i ("a initial") and event_page['time'][-1] as a_f ("a final") for a given EventPage a, if b follows a and then b_f >= a_i. In English, each EventPage's highest time must be greater than or equal to the preceding EventPages' lowest times.

PR #104, opened 2019-08-21
Include foreign key in exception.

danielballan:
At two places in the code, we can raise
``UnresolveableForeignKeyError``. One is when we encounter a datum_id
that we don't recognize and the other is when we encounter a
resource_uid that we don't recognize.

Currently, in the first case, we monkey-patch the relevant key onto the
exception, but in the second case, we don't.

In this PR, the second case includes the key too. And, in both cases, we
avoid monkey-patching and instead pass the relevant key to the
``__init__`` of ``UnresolveableForeignKeyError``. This is motivated by
[examples in the Python documentation](https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions)
showing that it's OK for user-defined exceptions to have custom
signatures, which I was not sure of when I wrote the first verison of
this code.

This PR also includes some minor unrelated flake8 fixes that came up in
a recent upgrade to flake8.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -533,7 +533,8 @@ class Filler(DocumentRouter):
                 try:
                     datum_doc = self._datum_cache[datum_id]
                 except KeyError as err:
-                    err_with_key = UnresolvableForeignKeyError(datum_id,
+                    err_with_key = UnresolvableForeignKeyError(
+                        datum_id,
                         f"Event with uid {doc['uid']} refers to unknown Datum "
                         f"datum_id {datum_id}")
                     raise err_with_key from err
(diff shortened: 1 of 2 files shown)


event_model/__init__.py
```diff
@@ -534,16 +534,17 @@ def fill_event(self, doc, include=None, exclude=None, inplace=None):
                     datum_doc = self._datum_cache[datum_id]
                 except KeyError as err:
                     err_with_key = UnresolvableForeignKeyError(
```
awalter-bnl: not a merge stopping comment but is there a reason why you didnt do just
``raise UnresolvableForeignKey(datum_id, ...)``

as you did below? it is probably a little neater

danielballan: Not sure what I was thinking. Thanks for pointing it out. Will push a fix.

PR #115, opened 2019-10-09
Fix handler_cache management.

### LINKED ISSUE #114: Clear handler_cache when the handler class for a spec is changed.
While we were debugging a beamline issue together, @mrakitin pointed out that registering a handler for some spec should cause any entries in the `Filler._handler_cache` associated with that spec to be cleared. Otherwise, if the user tried to update the `Filler.handler_registry` and then re-access data, they will confusingly be handed instances of the old class.

One way to implement this would be to make `handler_registry`, which is currently just a `dict`, a special `dict` subclass that can do extra things on `__setitem__`.

danielballan:
Closes #114.

~Tests currently fail, just pushed my work in progress so we don't lose it.~

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -824,6 +824,7 @@ class UnresolvableForeignKeyError(EventModelValueError):
         self.key = key
         self.message = message
 
+
 class DuplicateHandler(EventModelRuntimeError):
     "raised when a handler is already registered for a given spec"
     ...
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -775,6 +775,13 @@ class RunRouter(DocumentRouter):
 class EventModelError(Exception):
     ...
 
+# Here we define subclasses of all of the built-in Python exception types (as
+# needed, not a comprehensive list) so that all errors raised *directly* by
+# event_model also inhereit from EventModelError as well as the appropriate
+# built-in type. This means, for example, that `EventModelValueError` can be
+# caught by `except ValueError:` or by `except EventModelError:`. This can be
+# useful for higher-level libraries and for debugging.
+
 
 class EventModelKeyError(EventModelError, KeyError):
     ...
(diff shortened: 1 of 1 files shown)


event_model/__init__.py
```diff
@@ -453,6 +480,29 @@ def get_default_handler_cache():
     def inplace(self):
         return self._inplace
 
+    def register_handler(self, spec, handler, overwrite=False):
+        if (not overwrite) and (spec in self._handler_registry):
+            original = self._handler_registry[spec]
+            if original is handler:
+                return
+            raise DuplicateHandler(
+                f"There is already a handler registered for the spec {spec!r}."
```
```suggestion
                f"There is already a handler registered for the spec {spec!r}. "
```

event_model/__init__.py
```diff
@@ -673,6 +728,10 @@ def __init__(self, key, message):
         self.key = key
         self.message = message
 
+class DuplicateHandler(EventModelRuntimeError):
+    "raised when a handler is already registered for a given spec"
```
```suggestion
    "Raised when a handler is already registered for a given spec."
```

event_model/__init__.py
```diff
@@ -648,6 +699,10 @@ class EventModelRuntimeError(EventModelError, RuntimeError):
     ...
 
 
+class EventModelTypeError(EventModelError, TypeError):
+    ...
```
mrakitin: Do you mind to add a docstring here as well to have a better understanding of the purpose of this exception?

mrakitin: I think it's covered by 3ea1e63.

danielballan: Good point. If you look above and below you'll see this is one of many similar exceptions. I've added a comment in 3ea1e63b167cc0bc1cc8f6e54d79e153420449e7 to explain the motivation for all.

mrakitin: OK, sounds reasonable.

event_model/__init__.py
danielballan: Hmm, I had meant to commit this. Applied manually.

mrakitin: No worries, glad it's all in.

PR #116, opened 2019-10-11
Add option for RunRouter to fill documents.

danielballan:
In several instances (various demos and some beamlines) we have worked
out contorted code to handle the pattern:

```
RunEngine ---> (maybe 0MQ...) --> Filler ---> suitcase
```

where the consumer could be a suitcase or really anyconsumer that
requires filled data, like a live image viewer. There is currently no
elegant way to do this because `Filler` is a sink: there is no way to
"subscribe" to it.

One possible solution is to add `Filler.subscribe`, but another, which
is better for resource management and also simpler for users, is to put
`RunRouter` in charge of creating and destroying `Filler` instances.
That's what this PR does.

Notice that it provides hooks to:

* Just provide a `handler_registry` (the simple, common case)
* Decide whether to fill only certain specs---ones that aren't in the
  registry will be ignored by default
* Customize the type of Filler (e.g. could be a `DaskFiller`)
* Provide custom caches to Filler via passing in `filler_class=functools.partial(Filler ...)`

This was worked out in conversation with @tacaswell.

This needs tests.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -166,8 +166,13 @@ class RunRouter(DocumentRouter):
         containing callbacks that will receive all Events that reference that
         EventDescriptor and finally the RunStop document for the run.
     """
-    def __init__(self, factories):
+    def __init__(self, factories, handler_registry=None, *,
+                 root_map=None, filler_class=Filler, fill_or_fail=False):
         self.factories = factories
+        self.handler_registry = handler_registry or {}
+        self.filler_class = filler_class
+        self.fill_or_fail = fill_or_fail
+        self.root_map = root_map
 
         # Map RunStart UID to "subfactory" functions that want all
         # EventDescriptors from that run.
@@ -193,11 +198,18 @@ class RunRouter(DocumentRouter):
         # facilitate efficient cleanup of the caches above.
         self._descriptors = defaultdict(list)
 
+        # Map EventDescriptor UID to RunStart UID. This is used for looking up
+        # Fillers.
+        self._descriptor_to_start = {}
+
         # Map Resource UID to RunStart UID.
         self._resources = {}
 
         # Old-style Resources that do not have a RunStart UID
         self._unlabeled_resources = deque(maxlen=10000)
+
+        # Map Runstart UID to instances of self.filler_class.
+        self._fillers = {}
 
     def __repr__(self):
         return (f"RunRouter([\n" +
@@ -206,6 +218,12 @@ class RunRouter(DocumentRouter):
 
     def start(self, doc):
         uid = doc['uid']
+        filler = self.filler_class(self.handler_registry,
+                                   root_map=self.root_map,
+                                   inplace=False)
+        self._fillers[uid] = filler
+        # No need to pass the document to filler
+        # because Fillers do nothing with 'start'.
         for factory in self.factories:
             callbacks, subfactories = factory('start', doc)
             self._factory_cbs_by_start[uid].extend(callbacks)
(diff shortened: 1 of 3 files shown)


danielballan:
Tests added. And here's an interactive "test" / demo:

```py
In [1]: from bluesky import RunEngine                                                                                                                                                         

In [2]: from bluesky.plans import count                                                                                                                                                       

In [3]: from ophyd.sim import NumpySeqHandler, img                                                                                                                                            

In [4]: from event_model import RunRouter                                                                                                                                                     

In [5]: RE = RunEngine({})                                                                                                                                                                    

In [6]: from suitcase.tiff_stack import Serializer                                                                                                                                            

In [7]: def factory(name, doc): 
   ...:     serializer = Serializer('/tmp/test_run_router_fills') 
   ...:     serializer(name, doc) 
   ...:     return [serializer], [] 
   ...:                                                                                                                                                                                       

In [8]: rr = RunRouter([factory], {'NPY_SEQ': NumpySeqHandler})                                                                                                                               

In [9]: RE.subscribe(rr)                                                                                                                                                                      
Out[9]: 0

In [10]: RE(count([img]))                                                                                                                                                                     
Out[10]: ('9da4b611-c89b-40df-a7a6-ec1e766f2bf4',)

In [11]: ls /tmp/test_run_router_fills/                                                                                                                                                       
9da4b611-c89b-40df-a7a6-ec1e766f2bf4-primary-img.tiff
```

danielballan:
I had to rearrange the order that `RunRouter` and `Filler` are defined in the module in order for `RunRouter` to use `Filler` is a default value for a parameter. This rearrangement makes it hard to see the actual changes in the full diff. Reviewing [this specific commit](https://github.com/bluesky/event-model/pull/116/commits/fdca33e89329944ca9aed2735aee7f39fd984368) will make it easier to see the changes.

event_model/__init__.py
```diff
@@ -632,6 +457,250 @@ def __call__(self, name, doc, validate=False):
...
+        self._descriptors[start_uid].append(uid)
+        # Keep track of the EventDescriptor UID -> RunStartUID for filling
+        # purposes.
+        self._descriptor_to_start[uid] = start_uid
+
+    def event_page(self, doc):
+        descriptor_uid = doc['descriptor']
+        start_uid = self._descriptor_to_start[descriptor_uid]
+        try:
+            doc = self._fillers[start_uid].event_page(doc)
```
gwbischof: This line can also raise UnresolvableForeignKeyError. Should we handle this exception also?

gwbischof:
What is the mechanism to use RunRouter without filling? The previous RunRouter didn't fill, and the new RunRouter fills by default. Maybe we shouldn't make RunRouter fill by default so it doesn't break existing code?

danielballan:
Thanks for keeping an eye on backward-compatibility. True, the new RunRouter fills by default _if you give it a handler registry_. But the default value of `handler_registry` is `{}` and the default value of `fill_or_fail` is `False`, so the effect is that it still does not do any filling by default.

event_model/__init__.py
danielballan: Good question. I think cannot ignore that exception, given when it is raised. Suppose that I make `RunRouter([...], handler_registry={'SOME_SPEC': ...})`. If I then give that RunRouter an Event that references some `datum_id` that it doesn't recognize, it raise `UnresolvableForeignKeyError`. At that moment, I don't yet know whether the data relates to `'SOME_SPEC'` or not because I haven't gotten to the step where we have the Resource yet. Since the user is expecting all data from `'SOME_SPEC'` to be filled, I can't ignore the `UnresolvableForeignKeyError` and let the data pass through unfilled.

So, in effect, this PR is making the RunRouter stricter: it's adding validation. But the constraint that it is enforcing is one that we should be enforcing.

PR #120, opened 2019-10-18
If Resource root was change, note that.

danielballan:
If handler creation fails because a file is not found,
the error message can be confusing because it shows the Resource
document but does not include any information about whether the
'root' was overridden by ``root_map``.

The phrasing here could probably be improved upon. Any ideas?

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -556,8 +556,8 @@ class Filler(DocumentRouter):
                 f"handler registry.") from err
         # Apply root_map.
         resource_path = resource['resource_path']
-        root = resource.get('root', '')
-        root = self.root_map.get(root, root)
+        original_root = resource.get('root', '')
+        root = self.root_map.get(original_root, original_root)
         if root:
             resource_path = os.path.join(root, resource_path)
         try:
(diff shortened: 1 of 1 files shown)


tacaswell:
We should probably mention the existence of root map and the value of root in all cases.  I suspect we are more likely going to have issues because things are not in the root map rather than because they are erroneously being mapped.

danielballan:
Good suggestion. I now report the `root_map` in ever case and, if applicable report it before/after the `root_map` change.

event_model/__init__.py
```diff
@@ -566,10 +566,13 @@ def get_handler(self, resource):
         except Exception as err:
             msg = (f"Error instantiating handler "
                    f"class {handler_class} "
-                   f"with Resource document {resource}.")
-            if root != resouce.get('root', ''):
-                msg += (f" Note that its 'root' field was "
-                        f"mapped to {root} by root_map.")
+                   f"with Resource document {resource}. ")
+            if root != original_root:
+                msg += (f"It 'root' field was "
```
jklynch: "Its"?

PR #127, opened 2019-11-13
Port NoFiller from databroker.

danielballan:


danielballan:
Thanks for the nudge, @mrakitin. This was a good call.

PR #129, opened 2019-12-03
Factor dispatch logic out of call.

danielballan:
This makes it easier for subclasses to modfiy `__call__` but reuse the
dispatch logic. Useful in particular for making `QtAwareCallback`
better.

Transcribed from a snippet by @tacaswell

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -71,6 +71,14 @@ class DocumentRouter:
             The same name as what was passed in, and a doc that may be the same
             instance as doc, a copy of doc, or a different dict altogether.
         """
+        return self._dispatch(name, doc, validate)
+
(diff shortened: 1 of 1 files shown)


event_model/__init__.py
```diff
@@ -71,6 +71,14 @@ def __call__(self, name, doc, validate=False):
             The same name as what was passed in, and a doc that may be the same
             instance as doc, a copy of doc, or a different dict altogether.
         """
+        return self._dispatch(name, doc, validate)
+
+    def _dispatch(self, name, doc, validate):
+        """
+        Dispatch to the method corresponding to the `name`.
```
```suggestion
        Dispatch to the method corresponding to the ``name``.
```

danielballan: This method is private and won't show up in sphinx. :-D

PR #131, opened 2019-12-11
Allow DocRouter subclass to define page or non-page.

danielballan:
Currently, `DocumenterRouter` routes any Events it sees to the
`event_page` method, packing them into EventPages along the way. This
means that if someone subclasses `DocumentRouter` and does something
useful the `event_page` method, it will apply to `event` as well,
automatically.

However, the reverse does not work. If a subclass defines `event` but
not `event_page`, any Events will be handled but EventPages will pass
through quietly.

This PR allows a `DocumentRouter` subclass to define `event` *or*
`event_page` and have the documents flow to the one that is defined,
working in the same way that `__add__` and `__radd__` do in Python. As
before, the subclass may define both if it wishes (potentially useful
for performance optimizations) or neither.

This PR needs careful tests.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -97,8 +97,6 @@ class DocumentRouter:
                 # DatumPage or None or NotImplemented.
                 output_datum_page = self.datum_page(datum_page) or datum_page
                 if output_datum_page is not NotImplemented:
-                    if output_datum_page is None:
-                        output_datum_page = doc
                     output_doc, = unpack_datum_page(output_datum_page)
             elif name == 'event_page':
                 output_events = []
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -81,8 +81,8 @@ class DocumentRouter:
         """
         output_doc = getattr(self, name)(doc)
 
-        # If 'event' is not defined by the subclass by 'event_page' is, or vice
-        # versa, use that. And the same for 'datum_page' / 'datum.
+        # If 'event' is not defined by the subclass but 'event_page' is, or
+        # vice versa, use that. And the same for 'datum_page' / 'datum.
         if output_doc is NotImplemented:
             if name == 'event':
                 event_page = pack_event_page(doc)
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -93,7 +93,7 @@ class DocumentRouter:
                     output_doc, = unpack_event_page(output_event_page)
             elif name == 'datum':
                 datum_page = pack_datum_page(doc)
-                # Subclass' implementation of event_page may return a valid
+                # Subclass' implementation of datum_page may return a valid
                 # DatumPage or None or NotImplemented.
                 output_datum_page = self.datum_page(datum_page) or datum_page
                 if output_datum_page is not NotImplemented:
@@ -103,7 +103,7 @@ class DocumentRouter:
             elif name == 'event_page':
                 output_events = []
                 for event in unpack_event_page(doc):
-                    # Subclass' implementation of event_page may return a valid
+                    # Subclass' implementation of event may return a valid
                     # Event or None or NotImplemented.
                     output_event = self.event(event) or event
                     if output_event is NotImplemented:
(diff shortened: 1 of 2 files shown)


event_model/__init__.py
```diff
@@ -80,47 +80,73 @@ def _dispatch(self, name, doc, validate):
...
         output_doc = getattr(self, name)(doc)
+
+        # If 'event' is not defined by the subclass but 'event_page' is, or
+        # vice versa, use that. And the same for 'datum_page' / 'datum.
+        if output_doc is NotImplemented:
+            if name == 'event':
+                event_page = pack_event_page(doc)
+                output_event_page = self.event_page(event_page) or event_page
+                # Subclass' implementation of event_page may return a valid EventPage
+                # or None.
```
gwbischof: or NotImplemented

event_model/__init__.py
```diff
@@ -80,47 +80,79 @@ def _dispatch(self, name, doc, validate):
...
+                output_event_page = self.event_page(event_page) or event_page
+                if output_event_page is not NotImplemented:
+                    output_doc, = unpack_event_page(output_event_page)
+            elif name == 'datum':
+                datum_page = pack_datum_page(doc)
+                # Subclass' implementation of datum_page may return a valid
+                # DatumPage or None or NotImplemented.
+                output_datum_page = self.datum_page(datum_page) or datum_page
+                if output_datum_page is not NotImplemented:
+                    if output_datum_page is None:
```
gwbischof: This line doesn't appear in the if name == 'event' case

danielballan: Thanks. Note to self: the fix is to remove it here. Using `or datum_page` above handles it and guarantees that `output_datum_page` will not be `None`.

event_model/__init__.py
```diff
@@ -80,47 +80,79 @@ def _dispatch(self, name, doc, validate):
...
+                        output_datum_page = doc
+                    output_doc, = unpack_datum_page(output_datum_page)
+            elif name == 'event_page':
+                output_events = []
+                for event in unpack_event_page(doc):
+                    # Subclass' implementation of event may return a valid
+                    # Event or None or NotImplemented.
+                    output_event = self.event(event) or event
+                    if output_event is NotImplemented:
+                        break
```
mrakitin: What happens if `output_event` is None?

danielballan: This line

```py
output_event = self.event(event) or event
```

ensures that if `self.event(event)` returns `None` (which is quite possible) then `output_event` will be set to `event`, which is the document originally passed in to the DocumentRouter. We assume that this is a valid Event document, never `None`. (And, if validation is turned on, we verify that assumption.)

tacaswell:
Would it be simpler to not define the dispatched-to methods and watch for `AttributeError`?

danielballan:
I think `NotImplemented` is a language feature that probably should be used more than it is. An `AttributeError` can mean "Um, what? Do you have the wrong object?" `NotImplemented` means "Reasonable question! But you are barking up the wrong tree." It's more explicit. @tacaswell also pointed out (via DM) that the `NotImplemented` involves less stack thrashing and allows child classes  to "turn off" a parent class' implementation.

danielballan:
Added tests.

event_model/tests/test_em.py
```diff
@@ -340,6 +340,154 @@ def test_document_router_():
...
+            assert doc['data']['x'][0] == 1
+            mutable.append(object())
+
+    dr = DefinesEventPageAndEvent()
+    # Test that Event is routed to Event.
+    dr('event', event1)
+    assert len(mutable) == 1
+    mutable.clear()
+    # Test that EventPage is routed to EventPage.
+    dr('event_page', event_page)
```
gwbischof: Maybe there should be two separate mutable lists? One that gets appended to when event gets called, and another when event_page gets called.  Without two separate lists I don't think we have proven that the correct method was called.

event_model/tests/test_em.py
```diff
@@ -340,6 +340,154 @@ def test_document_router_():
...
+
+    dr = DefinesDatumPageAndDatum()
+    # Test that Datum is routed to Datum.
+    dr('datum', datum1)
+    assert len(mutable) == 1
+    mutable.clear()
+    # Test that DatumPage is routed to DatumPage.
+    dr('datum_page', datum_page)
+    assert len(mutable) == 1
+    mutable.clear()
```
gwbischof: Same as above. I think that we need two separate mutable lists to prove that the correct method has been called.

PR #133, opened 2019-12-20
Better error when packing empty page.

danielballan:
Currently, calling `pack_event_page()` with no arguments hits an
unexpected code path and thus fails like

```
UnboundLocalError: local variable 'event' referenced before assignment
```

This PR causes it to raise with a more describing `ValueError`. It is
not possible to pack an "empty" Event Page because an Event Page must
have a time and a uid.

The situation in the same for Datum Page, which must have a Resource.
That is also addressed here.

Tests included, should be good to merge.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1393,7 +1393,7 @@ def pack_datum_page(*datum):
         raise ValueError(
             "The pack_datum_page() function was called with empty *args. "
             "Cannot create an DatumPage from an empty collection of Datum "
-            "because the 'resource' field in an DatumPage cannot be NULL.")
+            "because the 'resource' field in a DatumPage cannot be NULL.")
     datum_id_list = []
     datum_kwarg_list = []
     for datum_ in datum:
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1322,7 +1322,10 @@ def pack_event_page(*events):
     event_page : dict
     """
     if not events:
-        raise ValueError("Cannot pack an event_page with no events.")
+        raise ValueError(
+            "The pack_event_page() function was called with empty *args. "
+            "Cannot create an EventPage from an empty collection of Events "
+            "because the 'descriptor' field in an EventPage cannot be NULL.")
     time_list = []
     uid_list = []
     seq_num_list = []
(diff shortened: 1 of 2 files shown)


tacaswell:
Per discussion on slack, the problem here isn't that we would have empty uids and times, it is that without at least one `Event` we can not infer the descriptor,   To quote @danielballan 

> In summary, an empty EventPage is OK, but an EventPage will a NULL descriptor field is not OK.

I have a very weak preference that the wording in the exceptions be changed to reflect this, but will not block merging on that.

event_model/__init__.py
```diff
@@ -1384,6 +1389,11 @@ def pack_datum_page(*datum):
     -------
     datum_page : dict
     """
+    if not datum:
+        raise ValueError(
+            "The pack_datum_page() function was called with empty *args. "
+            "Cannot create an DatumPage from an empty collection of Datum "
+            "because the 'resource' field in an DatumPage cannot be NULL.")
```
```suggestion
            "because the 'resource' field in a DatumPage cannot be NULL.")
```

event_model/__init__.py
```diff
@@ -1384,6 +1389,11 @@ def pack_datum_page(*datum):
     -------
     datum_page : dict
     """
+    if not datum:
+        raise ValueError(
+            "The pack_datum_page() function was called with empty *args. "
+            "Cannot create an DatumPage from an empty collection of Datum "
```
```suggestion
            "Cannot create an DatumPage from an empty collection of Datums "
```

danielballan: "Datums" makes me cringe since the English (Latin, really) plural of "datum" is "data".

event_model/__init__.py
danielballan: "Datums" makes me cringe since the English (Latin, really) plural of "datum" is "data".

event_model/__init__.py
danielballan: "Datums" makes me cringe since the English (Latin, really) plural of "datum" is "data".

mrakitin: I agree, thought about it, but wanted to make it consistent with the `EventPages` above. Let's leave it as is here.

PR #136, opened 2020-01-28
Documentation on external assets

danielballan:
This is a start at some much-needed documentation on how external assets
work in bluesky. This is developer documentation: more practical
guidance belongs in ophyd (the creation side) and databroker (the
retrieval side). It is intended to explain the motivation for and
details of how the relevant information is organized in documents, how
the handler API works, how handler discovery works, and how the new
Filler and RunRouter fit in.

-->


docs/source/data-model.rst
```diff
@@ -274,7 +274,23 @@ Formal schema:
 Event contents can also be represented in "paged" form, where multiple
 rows are contained in one structure for efficient transport and vectorized
 computation. The representations contain equivalent information: an EventPage
-can always be losslessly transformed into an Event and vice versa.
+can always be losslessly transformed into an Event and vice versa. Here is the
```
prjemian: ... can always be transformed, without loss, into ...

danielballan: Done. Always good to avoid needless jargon when we  can.

docs/source/data-model.rst
```diff
@@ -322,6 +350,18 @@ Minimal nontrivial valid example:
     'spec': 'SOME_SPEC',
     'uid': '272132cf-564f-428f-bf6b-149ee4287024'}
 
+Typical example:
```
tacaswell: Do we want to delete the version just above?

docs/source/external.rst
```diff
@@ -0,0 +1,297 @@
...
+
+   # 'event' document
+   {'data': {'image': 'aa10035d-1d2b-41d9-97e6-03e3fe62fa6c/5',
+             'temperature': 5.0},
+    'descriptor': '219310e0-faa0-4990-84a0-95b508d4ae35',
+    ...}
+
+where the other fields in the Event have been omitted (``...``) for brevity.
+We can tell that the value of ``'image'`` is a placeholder, a foreign key
+referencing some array yet to be retrieved, by consulting the Event Descriptor
```
tacaswell:
```suggestion
referencing some array yet to be retrieved, by consulting the Event Descriptor
```
```suggestion
referencing some array yet to be retrieved, by consulting the Datum
```

danielballan:
> I'm wondering about the representation of a scalar value as a vector of length 1. Do we need to make these the same, thereby losing the clear distinction between both?

I agree that we want a clear distinction between these. I think we have it. In the event page

```
   # 'event_page' document
   {'data': {'random_walk:dt': [-1.0],
             'random_walk:x': [1.9221013521832928]},
   <snipped>
```

think of the value of `'data'` as a DataFrame: a dict of lists/arrays. The fact that `len(event_page['data']['random_walk:x'] == 1` tells us only that this is a length-1 Event Page. It doesn't tell us anything about the shape of the data contained therein. To get that, we might look at the number of dimensions in the first row of the page, `numpy.ndim(event_page['data']['random_walk:x'][0])` which would correctly return `0`. If we meant to represent a length-1 scalar there, we would have

```
   # 'event_page' document
   {'data': {'random_walk:dt': [[-1.0]],
             'random_walk:x': [[1.9221013521832928]]},
   <snipped>
```

That is, ``[[-1.0]]` in place of `[-1.0]`.

docs/source/data-model.rst
danielballan: I see your point, but we have "Minimal" and "Typical" examples in every other case, and it might help to illustrate that `resource_kwargs` has not requires keys.

docs/source/external.rst
```diff
@@ -0,0 +1,297 @@
...
+typically implemented using a class that implements ``__init__`` and
+``__call__``, with the respective signatures.
+
+.. code:: python
+
+   class MyHandler:
+       def __init__(self, path, **resource_kwargs):
+           # Consume the path information and the 'resource_kwargs' from the
+           # Resource. Typically stashes some state and/or opens file(s).
+           ...
```
ronpandolfi: I wonder if it would help to describe what the expected lifecycle of a handler is? Does DB cache the handler instance for consecutive reads? How does that cache expire?

tacaswell: That is an implementation detail we have struggled with and one of the things that motivated the design of RunRouter to scope the life to "between the start and stop documents" which is commented on later.

IRRC, the cache is injectable so it is also possible to have very custom policies on a per-application basis.

danielballan: Seconded. The Filler and RunRouter classes make it easier to clean up correctly and set custom cache policies, but I still feel we haven’t settled on what the right policies should be for DataBroker.

There are some comments on resource management later on in the document. I will add some more here.

danielballan: Added comments in https://github.com/bluesky/event-model/pull/136/commits/cf4d0097085b435fb52acf1aa0e8ee890aca16e4. Any thoughts?

ronpandolfi: Very helpful explanation; the flexibility is interesting. Thanks!

danielballan:
Having addressed @prjemian comments and received two approvals, I'll merge to get this published, but comments from @dylanmcreynolds, @mrakitin, @jklynch, or anyone else are very much welcome.

PR #137, opened 2020-01-28
Release notes for v1.13.1

danielballan:


danielballan:
Forgive for taking the first approval and running with this so I can get the tag out. Happy to make revisions after the fact as needed, @jklynch.

PR #139, opened 2020-01-30
Fix critical typo.

danielballan:
Fixes a bug found at SST. Not sure how to test this yet, but we probably
should before merging.


danielballan:
Added a test in 9638bb1. Confirmed that it fails without the fix in 7b0ffc5.

PR #140, opened 2020-02-01
DOC: tiny fixes after #136 review

mrakitin:
Small corrections for the merged #136.


danielballan:
Wouldn’t it be “an LRU cache”?

danielballan:
Oh I see: it renders as “a cachetools LRU cache”. Good catch.

PR #141, opened 2020-02-04
Document rule against structure in Event data values

danielballan:
This documents a design decision that one might be able to infer from the schema but should also be stated and justified explicitly.

I also added some cross-references following up on #136.


tacaswell:
A bit of reference, if you allow 1 level of nesting in `data`, then you quickly end up noting that the restriction _those_ values should allow nesting and so on which leads you to accepting arbitrarily nested structured data.  This in turn makes the descriptions significantly more complex.

danielballan:
Thanks, @tacaswell. That's a good point and seems worth including. Added (paraphrased) in 39f3bca.

PR #143, opened 2020-02-10
Add an example of Filler cache management.

danielballan:
This adds one example to give a sense of how Filler cache management can be customized. I like this addition, and I don't think it interrupts the flow of the narrative too much.


docs/source/external.rst
```diff
@@ -262,7 +262,23 @@ object, such as a :class:`cachetools.LRUCache` or
 :class:`cachetools.LFUCache`, to receive a prepopulated cache, or to share
 caches between Filler instances. This is an implementation detail left entirely
 up to the application. See :class:`~event_model.Filler` for details on cache
-injection.
+injection. Here is an example where two Fillers share a global LRU cache:
+
+.. code:: python
+
+   import event_model
+   import cache_tools
```
mrakitin:
```suggestion
   import cachetools
```

PR #147, opened 2020-02-22
Error message formatting tweak.

danielballan:
Put quotes around string.


danielballan:
Self-merging as trivial.

PR #152, opened 2020-02-25
Added InvalidData and MismatchedDataKeys exceptions

### LINKED ISSUE #151: Add an InvalidDocuments exception
Durring the testing of databroker on NSLS2 data we found that occasionally we get a KeyError when filling an event because event['data'].keys() != descriptor['data_keys'].keys(). This is invalid.

I think this we should add an InvalidDocuments exception, so that we have a more meaningful explanation of the problem.

```    
def fill_event(self, doc, include=None, exclude=None, inplace=None):
        try:
            filled = doc['filled']
        except KeyError:
            # This document is not telling us which, if any, keys are filled.
            # Infer that none of the external data is filled.
            descriptor = self._descriptor_cache[doc['descriptor']]
            filled = {key: 'external' in val
                      for key, val in descriptor['data_keys'].items()}
        for key, is_filled in filled.items():
            if exclude is not None and key in exclude:
                continue
            if include is not None and key not in include:
                continue
            if not is_filled:
>               datum_id = doc['data'][key]
E               KeyError: 'Synced_saxs_image'
```

gwbischof:
Fixes #151

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -645,8 +645,8 @@ class Filler(DocumentRouter):
             filled_doc = doc
         else:
             filled_doc = copy.deepcopy(doc)
-
         descriptor = self._descriptor_cache[doc['descriptor']]
+        from_datakeys = False
         self._current_state.descriptor = descriptor
         try:
             filled = doc['filled']
@@ -655,6 +655,7 @@ class Filler(DocumentRouter):
             # Infer that none of the external data is filled.
             filled = {key: 'external' in val
                       for key, val in descriptor['data_keys'].items()}
+            from_datakeys = True
         for key, is_filled in filled.items():
             self._current_state.key = key
             if exclude is not None and key in exclude:
@@ -662,7 +663,21 @@ class Filler(DocumentRouter):
             if include is not None and key not in include:
                 continue
             if not is_filled:
-                datum_id = doc['data'][key]
+                try:
+                    datum_id = doc['data'][key]
+                except KeyError as err:
+                    if from_datakeys:
+                        raise MismatchedDataKeys(
+￼                            "The documents are not valid.  Either because they "
+￼                            "were recorded incorrectly in the first place, "
+￼                            "corrupted since, or exercising a yet-undiscovered "
+￼                            "bug in reader. event['data'].keys()"
+￼                            "must equal descriptor['data_keys'].keys()."
+￼                            f"event['data'].keys(): {doc['data'].keys()}, "
+￼                            "descriptor['data_keys'].keys(): "
+￼                            f"{descriptor['data_keys'].keys()}") from err
+                    else:
+                        raise err
                 # Look up the cached Datum doc.
                 try:
                     datum_doc = self._datum_cache[datum_id]
@@ -784,13 +799,13 @@ class NoFiller(Filler):
         return filled_doc
 
     def fill_event(self, doc, include=None, exclude=None, inplace=None):
+        descriptor = self._descriptor_cache[doc['descriptor']]
+        from_datakeys = False
         try:
             filled = doc['filled']
-            from_datakeys = False
         except KeyError:
             # This document is not telling us which, if any, keys are filled.
             # Infer that none of the external data is filled.
-            descriptor = self._descriptor_cache[doc['descriptor']]
             filled = {key: 'external' in val
                       for key, val in descriptor['data_keys'].items()}
             from_datakeys = True
@@ -812,7 +827,7 @@ class NoFiller(Filler):
                             "must equal descriptor['data_keys'].keys()."
                             f"event['data'].keys(): {doc['data'].keys()}, "
                             "descriptor['data_keys'].keys(): "
-                            f"{self._descriptor_cache()}") from err
+                            f"{descriptor['data_keys'].keys()}") from err
                     else:
                         raise err
                 # Look up the cached Datum doc.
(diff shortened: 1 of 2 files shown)


event_model/__init__.py
```diff
@@ -786,19 +786,35 @@ def fill_event_page(self, doc, include=None, exclude=None):
...
-                datum_id = doc['data'][key]
+                try:
+                    datum_id = doc['data'][key]
+                except KeyError as err:
+                    if from_datakeys:
+                        raise MismatchedDataKeys(
+                            "The documents are not valid.  Either because they "
+                            "were recorded incorrectly in the first place, "
+                            "corrupted since, or exercising a yet-undiscovered "
+                            "bug in reader. event['data'].keys()"
```
danielballan:
```suggestion
                            "bug in a reader. event['data'].keys() "
```

event_model/__init__.py
```diff
@@ -786,19 +786,35 @@ def fill_event_page(self, doc, include=None, exclude=None):
...
+                            "must equal descriptor['data_keys'].keys()."
+                            f"event['data'].keys(): {doc['data'].keys()}, "
+                            "descriptor['data_keys'].keys(): "
+                            f"{self._descriptor_cache()}") from err
+                    else:
+                        raise err
```
danielballan: This case would mean that there are keys in `event['filled']` which are not in `event['data']`, which is another way in which the keys could be mismatched. Is it worth raising the same exception with different message text here? We haven't seen this failure mode yet, and I doubt we'd see it from _databroker_ (because Events at rest have no `'filled'` key) but it might come up in stream applications with Events over the wire, which can have a `'filled'` key.

event_model/__init__.py
```diff
@@ -786,19 +786,35 @@ def fill_event_page(self, doc, include=None, exclude=None):
...
+                    if from_datakeys:
+                        raise MismatchedDataKeys(
+                            "The documents are not valid.  Either because they "
+                            "were recorded incorrectly in the first place, "
+                            "corrupted since, or exercising a yet-undiscovered "
+                            "bug in reader. event['data'].keys()"
+                            "must equal descriptor['data_keys'].keys()."
+                            f"event['data'].keys(): {doc['data'].keys()}, "
+                            "descriptor['data_keys'].keys(): "
+                            f"{self._descriptor_cache()}") from err
```
danielballan: I think this was a typo, or maybe I'm not following.
```suggestion
                            f"{descriptor['data_keys'].keys()}") from err
```

event_model/__init__.py
```diff
@@ -786,19 +786,35 @@ def fill_event_page(self, doc, include=None, exclude=None):
...
+                try:
+                    datum_id = doc['data'][key]
+                except KeyError as err:
+                    if from_datakeys:
+                        raise MismatchedDataKeys(
+                            "The documents are not valid.  Either because they "
+                            "were recorded incorrectly in the first place, "
+                            "corrupted since, or exercising a yet-undiscovered "
+                            "bug in reader. event['data'].keys()"
+                            "must equal descriptor['data_keys'].keys()."
```
danielballan:
```suggestion
                            "must equal descriptor['data_keys'].keys(). "
```

event_model/__init__.py
```diff
@@ -1119,7 +1135,17 @@ def __init__(self, key, message):
...
+    ...
+
+
+class InvalidData(EventModelError):
+    """raised when the data is invalid"""
+    ...
+
+
+class MismatchedDataKeys(InvalidData):
+    """raised when event['data'].keys() != descriptor['data_keys'].keys()"""
```
danielballan: could also be when any of the data keys structures, which include `event['timestamps'].keys()` and `event['filled'].keys()`, are mutually out of sync

danielballan:
That's a good question. I think that consumers and readers should not perform any _proactive_ validation. If they hit an unrecoverable error, it's worth doing some extra investigation to provide a specific error message, as this PR does, but I think it is fair to assume that the documents they read are valid unless proven otherwise.

As I mentioned on Slack but will rehash here for the sake of a public record, I think we should validate when the documents are composed  (in `event_model.compose_run`) and _maybe_ when they are persisted (in the suitcases). But any validation work that we do when the documents are _accessed_ is needlessly repeated every time they are accessed, and the notice also comes too late. It's better to fail when the documents are first created and/or persisted because the user is immediately notified about the problem, and you only pay for the performance overhead once.

So, as far as the Filler and databroker are concerned, I think it's OK to have a situation where malformed documents lead to keys silently not being filled. As a separate matter, we should review our validation functions to ensure that they would flag malformed documents at compose/write time, and then we should comb the existing databases for irregularities and decide what to do about the invalid documents that we find.

gwbischof:
The old databroker iterates the event['data'].keys() instead of descriptor['data_keys'].keys() (the way it is done here in event model)

tacaswell:
The gut checks out.  Don't fully follow the last paragraph about the Filler though.

tacaswell:
I think the code would be simpler if we included the keys of the descriptor, the data, filled, (and the timestamps) with the same text rather than doing the book keeping to raise _just_ the rightly worded exception.   All we really know is that something that is consistent is not, but we can not really be sure which of the inconsistent things in right (and going with majority voting is not always correct).

event_model/__init__.py
```diff
@@ -655,14 +655,37 @@ def fill_event(self, doc, include=None, exclude=None, inplace=None):
             # Infer that none of the external data is filled.
             filled = {key: 'external' in val
```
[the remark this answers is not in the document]
danielballan: Good catch. Yes.

PR #155, opened 2020-03-04
Add replay utility.

danielballan:
This ports a utility that we used for the FXI experiment from [my gist](https://gist.github.com/danielballan/516e0b0183f424a9783a57b056ae516c) to event-model.

The test shows that this does not work with `event_page`. Need some thought.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1822,3 +1822,42 @@ class NumpyEncoder(json.JSONEncoder):
                 return obj.item()
             return obj.tolist()
         return json.JSONEncoder.default(self, obj)
+
+
+def replay(gen, callback, burst=False):
+    """
+    Emit documents to a callback with realistic time spacing.
+
+    Parameters
+    ----------
+    gen: iterable
+        Expected to yield (name, doc) pairs
+    callback: callable
+        Expected signature: callback(name, doc)
+    burst: bool, optional
+        If True, emit as fast as possible, ignoring timing. False by default.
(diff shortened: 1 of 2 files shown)


event_model/__init__.py
```diff
@@ -1822,3 +1822,42 @@ def default(self, obj):
...
+    Parameters
+    ----------
+    gen: iterable
+        Expected to yield (name, doc) pairs
+    callback: callable
+        Expected signature: callback(name, doc)
+    burst: bool, optional
+        If True, emit as fast as possible, ignoring timing. False by default.
+    """
+    DOCUMENTS_WITHOUT_A_TIME = {'datum', 'datum_page', 'resource'}
```
prjemian: Looks like a list.  Why the curly braces?

tacaswell: Micro-optimization to use a set instead of a list.  It make the init a bit slower, makes `x in DOCUMENTS_WITHOUT_A_TIME` a bit faster.   Removes the `[n]` access because there are no meaningful semantics in the order of these values.

tacaswell:
Hmm, maybe we should take a time window to batch stuff into, an N to batch into, or just always break the pages up into individual events?

danielballan:
We have code in databroker that joins events across streams, subdividing pages only as necessary to ensure monotonic time. I think it would take more than trivial effort to factor it out, but that seems like the right path here. We want that code to be available widely.

PR #158, opened 2020-03-05
RunRouter passes more docs to callbacks.

### LINKED ISSUE #149: Make RunRouter pass 'start' and 'descriptor' to callbacks
When constructing a callback in a factory function for the `RunRouter`, such as

```py
from event_model import DocumentRouter, RunRouter

class SomeCallback(DocumentRouter):
    def start(name, doc):
        self._start_doc = doc

    def event(self, doc):
        # Do something with self._start_doc.

def factory(name, doc):
    cb(name, doc)
    cb = SomeCallback()
    return [cb], []

rr = RunRouter([factory])
```

it is easy to forget `cb(name, doc)`, i.e.

```py
def factory(name, doc):
    cb = SomeCallback()
    return [cb], []
```

which typically leads to a secondary error (`AttributeError: 'SomeCallback' object has not attribute '_start_doc'` in this example) that does not make it at all clear that the mistake lies in `factory`.

The same sort of problem can occur for `descriptor` in RunRouter sub-factories.

I have made this mistake myself many times. Should we try to provide better errors? We could flip a flag when the start doc goes through and keep a cache of descriptor uids we have seen to provide better errors when we are missing a start or descriptor doc. It's not obvious to me where would be the best place to do it. Options:

* Put a check in `DocumentRouter.__call__`
* Put a check in `DocumentRouter._dispatch`.
* Put in a check in the base methods of `start` and `descriptor`. This is complicated by the fact that for some methods (`event`, `event_page`, `datum`, `datum_page`) subclasses _must not_ return `super()` because the base class returns `NotImplemented`. Thus, it be would be strange to advise, "Be sure to call `super()` in some document methods but not in others."

danielballan:
Closes #149. See that issue for discussion.

Needs tests before merging.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -872,6 +872,18 @@ class NoFiller(Filler):
         return doc
 
 
+DOCS_PASSED_IN_1_14_0_WARNING = (
+    "The callback {callback!r} raised {err!r} when "
+    "RunRouter passed it a {name!r} document. This is "
+    "probably because in earlier releases the RunRouter "
+    "expected its factory functions forward the 'start' "
+    "document, but starting in event-model 1.14.0 the "
+    "RunRouter passes in the document, causing the "
+    "callback to receive it twice and potentially raise "
+    "an error. Update the factory function. In a future "
+    "release this warning will become an error.")
+
+
 class RunRouter(DocumentRouter):
     """
     Routes documents, by run, to callbacks it creates from factory functions.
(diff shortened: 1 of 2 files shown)


event_model/__init__.py
```diff
@@ -1006,6 +1007,20 @@ def start(self, doc):
...
             callbacks, subfactories = factory('start', doc)
+            for callback in callbacks:
+                try:
+                    callback('start', doc)
+                except Exception as err:
+                    warnings.warn(
+                        f"The callback {callback} raised a {err} when "
+                        f"RunRouter passed it a 'start' document. This is "
+                        f"probably because in earlier releases the RunRouter "
+                        f"expected its factory functions forward the 'start' "
```
jklynch: ...factory functions "to" forward...

event_model/__init__.py
```diff
@@ -1023,6 +1038,33 @@ def descriptor(self, doc):
...
             self._subfactory_cbs_by_descriptor[uid].extend(callbacks)
+            for callback in callbacks:
+                try:
+                    callback('start', doc)
+                except Exception as err:
+                    warnings.warn(
+                        f"The callback {callback} raised a {err} when "
+                        f"RunRouter passed it a 'start' document. This is "
+                        f"probably because in earlier releases the RunRouter "
+                        f"expected its factory functions forward the 'start' "
```
jklynch: ...factory functions "to" forward...

event_model/__init__.py
```diff
@@ -1023,6 +1038,33 @@ def descriptor(self, doc):
...
+                        f"an error. Update the factory function. In a future "
+                        f"release this warning will become an error.")
+                try:
+                    callback('descriptor', doc)
+                except Exception:
+                    warnings.warn(
+                        f"The callback {callback} raised a {err} when "
+                        f"RunRouter passed it a 'descriptor' document. This is "
+                        f"probably because in earlier releases the RunRouter "
+                        f"expected its subfactory functions forward the 'descriptor' "
```
jklynch: ...subfactory functions "to" forward...

PR #161, opened 2020-03-27
Make Filler pickle-able.

danielballan:
When we added `threading.local()` to Filler to support the coersion
feature, we made it un-pickle-able. This adds a test to defend its
pickel-ability and restores it.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -442,6 +442,61 @@ class Filler(DocumentRouter):
         self.retry_intervals = retry_intervals
         self._closed = False
 
+    def __eq__(self, other):
+        return (
+            type(self) is type(other) and
+            self.inplace == other.inplace and
+            self._coerce == other._coerce and
+            self.include == other.include and
+            self.exclude == other.exclude and
+            self.root_map == other.root_map and
+            type(self._handler_cache) is type(other._handler_cache) and
+            type(self._resource_cache) is type(other._resource_cache) and
+            type(self._datum_cache) is type(other._datum_cache) and
+            type(self._descriptor_cache) is type(other._descriptor_cache) and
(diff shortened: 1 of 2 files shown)


klauer:
(Not sure if it's applicable here, but chiming in regardless: generally I've switched to implementing https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__ as it's more flexible.)

event_model/__init__.py
```diff
@@ -442,6 +442,61 @@ def __init__(self, handler_registry, *,
...
+            self.inplace == other.inplace and
+            self._coerce == other._coerce and
+            self.include == other.include and
+            self.exclude == other.exclude and
+            self.root_map == other.root_map and
+            type(self._handler_cache) is type(other._handler_cache) and
+            type(self._resource_cache) is type(other._resource_cache) and
+            type(self._datum_cache) is type(other._datum_cache) and
+            type(self._descriptor_cache) is type(other._descriptor_cache) and
+            self.retry_intervals == other.retry_intervals
```
MikeHart85: Would it be unreasonable to compare `self.__getstate__()` and `other.__getstate__()` instead, to avoid the two implementations going out of sync? 

There is a slight difference, in that you're only checking type and not content of the caches, but as you pointed out in Slack that difference is a bit arbitrary.

Self-counter-argument: The test would become somewhat circular.
Counter-self-counter-argument: The test implies this should be guaranteed anyway.

danielballan: Yeah, I agree there's an argument for taking path, but I do think that treating the cache content as an internal is better.

MikeHart85: Makes sense; everything looks good to me.

Issue #165, opened 2020-04-07
Add convenience methods to DocumentRouter


jklynch:
I propose some convenience methods for DocumentRouter along these lines:

- DocumentRouter.start_doc()
      return the start document or raise an Exception if the start document has not been seen yet
- DocumentRouter.descriptor_for_event(event_doc)
   return the descriptor document associated with the given event document from a dictionary of uid --> descriptor
- DocumentRouter.stream_name(event_doc)
      given an event document return its stream name by looking up the event's descriptor in a dictionary of uid --> descriptor

I find myself reimplementing this functionality virtually every time I subclass DocumentRouter.

tacaswell:
I'm sold on the first two if we already keep enough state to answer that question, a bit skeptical but still net positive if we have to add the state tracking.

The third seems likes `self.descriptor_for_event(event_doc)['name']` so neutral on it.

jklynch:
addressed by #166

Issue #164, opened 2020-04-07
Helpful warning makes development difficult


jklynch:
A helpful warning was added to RunRouter.start(...) when its behavior was changed to pass the start document to each factory. Exceptions resulting from the new calls to [callback('start', doc)](https://github.com/bluesky/event-model/blob/master/event_model/__init__.py#L1083) trigger the warning and are _squashed_.

As I work on a new DocumentRouter subclass I wish the legitimate exceptions I am generating in `start(...)` would propagate all the way out and whack the interpreter just like all the other exceptions I generate.

## Desired Behavior
I would like exceptions described above to be re-raised in addition to triggering the warning.

danielballan:
Yes, I think our attempt to provide a gradual migration path was well-intentioned but flawed.

To give users the chance to defer updating the factories, we _could_ add a kwarg to opt into the old behavior. I think given the low adoption of RunRouter so far we might instead just rush the deprecation cycle and release 1.15.0 with the change proposed here and no kwarg. Users who want the old behavior can stay on 1.14.x.

tacaswell:
I have also come to like envs for controlling this sort of behavior.  Very much like Python and gcc have options to turn warnings into errors, maybe we should have an ecosystem wide env (like `BLUESKY_CRANKY`, `BLUESKY_FUTURE`) that turns all deprecation warnings / sanding of edges into errors?

danielballan:
OK, I agree that using an env variable here would be better than adding a kwarg. But I also think that "neither" is fine here.

jklynch:
The issue I have with the current behavior is the assumption that an exception passing through these try/catch blocks is not a real exception and should be erased. I think this is "bad" behavior and should not be an optional.

tacaswell:
I agree in principle, but we also don't want to break working code with no warning / deprecation period.

jklynch:
addressed by #166

PR #166, opened 2020-04-07
Add convenience methods and new warning to DocumentRouter

jklynch:
This PR proposes two modifications to DocumentRouter:

- deliver a warning if a DocumentRouter processes the same start document twice consecutively to address #164

- add convenience methods to address #165 
  - `get_start()` returning the most recently processed start document
  - `get_descriptor(event_doc)` returning the descriptor document corresponding to the specified event document
  - `get_stream_name(event_doc)` returning the stream name for the specified event document; this seems like a minor convenience but it is the piece of information I very often want from a descriptor

The convenience methods are prefixed with `get_` since DocumentRouter already defines `start(name, doc)` and `descriptor(name, doc)` but some other naming scheme might be better.

I realize now that the `event_doc` parameter for `get_descriptor` and `get_stream_name` should probably just be `doc`.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -185,6 +185,7 @@ class SingleRunDocumentRouter(DocumentRouter):
     A DocumentRouter intended to process events from exactly one run.
     """
     def __init__(self):
+        super().__init__()
         self._start_doc = None
         self._descriptors = dict()
 
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -214,15 +214,17 @@ class SingleRunDocumentRouter(DocumentRouter):
             if self._start_doc is None:
                 self._start_doc = doc
             else:
-                raise EventModelError(
-                    'SingleRunDocumentRouter received a second start document: uid {doc["uid"]}'
+                raise EventModelValueError(
+                    f'SingleRunDocumentRouter associated with start document {self._start_doc["uid"]} '
+                    f'received a second start document with uid {doc["uid"]}'
                 )
         elif name == 'descriptor':
             if doc['run_start'] == self._start_doc['uid']:
                 self._descriptors[doc['uid']] = doc
             else:
-                raise EventModelError(
-                    'SingleRunDocumentRouter received a descriptor associated with a different start document'
+                raise EventModelValueError(
+                    f'SingleRunDocumentRouter associated with start document {self._start_doc["uid"]} '
+                    f'received a descriptor {doc["uid"]} associated with start document {doc["run_start"]}'
                 )
 
         return super().__call__(name, doc, validate)
@@ -254,10 +256,10 @@ class SingleRunDocumentRouter(DocumentRouter):
         descriptor document : dict
         """
         if 'descriptor' not in doc:
-            raise ValueError(f'document is not associated with a descriptor:\n{doc}')
+            raise EventModelValueError(f'document is not associated with a descriptor:\n{doc}')
         elif doc['descriptor'] not in self._descriptors:
-            raise EventModelError(
-                'SingleRunDocumentRouter has not processed a descriptor with uid {doc["descriptor"]}'
+            raise EventModelValueError(
+                f'SingleRunDocumentRouter has not processed a descriptor with uid {doc["descriptor"]}'
             )
 
         return self._descriptors[doc['descriptor']]
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -53,10 +53,6 @@ class DocumentRouter:
 
         (name, getattr(router, name)(doc))
     """
-    def __init__(self):
-        self._start_doc = None
-        self._descriptors = dict()
-
     def __call__(self, name, doc, validate=False):
         """
         Process a document.
@@ -75,27 +71,7 @@ class DocumentRouter:
             The same name as what was passed in, and a doc that may be the same
             instance as doc, a copy of doc, or a different dict altogether.
         """
-        if name == 'start':
-            if self._start_doc is None:
-                self._start_doc = doc
-            elif self._start_doc['uid'] == doc['uid']:
-                warnings.warn(f'DocumentRouter received the same start document twice: uid {doc["uid"]}')
-            else:
-                self._start_doc = doc
-                # ??? self._descriptors = dict()
-        elif name == 'descriptor':
-            self._descriptors[doc['uid']] = doc
-
         return self._dispatch(name, doc, validate)
-
-    def get_start(self):
-        return self._start_doc
-
-    def get_descriptor(self, event_doc):
-        return self._descriptors[event_doc['descriptor']]
-
-    def get_stream_name(self, event_doc):
-        return self.get_descriptor(event_doc)['name']
 
     def _dispatch(self, name, doc, validate):
         """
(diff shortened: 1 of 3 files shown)


event_model/__init__.py
```diff
@@ -71,8 +75,28 @@ def __call__(self, name, doc, validate=False):
             The same name as what was passed in, and a doc that may be the same
             instance as doc, a copy of doc, or a different dict altogether.
         """
+        if name == 'start':
+            if self._start_doc is None:
+                self._start_doc = doc
+            elif self._start_doc['uid'] == doc['uid']:
+                warnings.warn(f'DocumentRouter received the same start document twice: uid {doc["uid"]}')
```
tacaswell, line 82: maybe not process it the second time?

tacaswell:
Needs tests and docstrings, but I'm :+1: on this. 

This is very common book keeping and in the grand scheme of things relatively cheap.  I'm still meh on the stream name helper, but won't argue over it.

event_model/__init__.py
jklynch: That was my first thought but the tests for Filler fail so I punted.

jklynch:
Here is a proposal for a minimal change to RunRouter. The warning is still issued but the exception is re-raised.

event_model/__init__.py
```diff
@@ -180,6 +180,103 @@ def bulk_datum(self, doc):
         self.datum_page(bulk_datum_to_datum_page(doc))
 
 
+class SingleRunDocumentRouter(DocumentRouter):
+    """
+    A DocumentRouter intended to process events from exactly
```
danielballan, line 185: Sphinx gives special significance to the first line of the docstring, for display purposes. Can this be one line?

event_model/__init__.py
```diff
@@ -180,6 +180,103 @@ def bulk_datum(self, doc):
...
+    one run.
+    """
+    def __init__(self):
+        self._start_doc = None
+        self._descriptors = dict()
```
danielballan, line 190: If we call `super().__init__()` here we can future-proof ourselves against possible changes to `DocumentRouter`.

event_model/__init__.py
```diff
@@ -180,6 +180,103 @@ def bulk_datum(self, doc):
...
+        Parameters
+        ----------
+        doc : dict
+            event-model document
+
+        Returns
+        -------
+        descriptor document : dict
+        """
+        if 'descriptor' not in doc:
```
danielballan, line 256: Micro-optimization: it is faster to "beg forgiveness" than "ask permission".

```
try:
    descriptor_uid = doc['descriptor']
except KeyError:
    ...
try:
    self._descriptors[descriptor_uid]
except KeyError:
    ...
```

event_model/__init__.py
```diff
@@ -180,6 +180,103 @@ def bulk_datum(self, doc):
...
+            raise ValueError(f'document is not associated with a descriptor:\n{doc}')
```
danielballan, line 257: All errors directly raised from inside `event_model` should be subclasses of `EventModelError`. This one could be `EventModelValueError`, which I believe is already defined below.

This someone picky pattern is one that we apply in event-model and the suitcase libraries, but have not taken the trouble to retroactively apply to the older libraries. The purpose is to make it clear whether event-model did a thing that cause _another_ library to raise (including a standard Python library) or whether the problem is directly inside event-model

event_model/__init__.py
```diff
@@ -180,6 +180,103 @@ def bulk_datum(self, doc):
...
+            else:
+                raise EventModelError(
+                    'SingleRunDocumentRouter received a second start document: uid {doc["uid"]}'
+                )
+        elif name == 'descriptor':
+            if doc['run_start'] == self._start_doc['uid']:
+                self._descriptors[doc['uid']] = doc
+            else:
+                raise EventModelError(
+                    'SingleRunDocumentRouter received a descriptor associated with a different start document'
```
danielballan, line 225: Including uids in here might be helpful.

event_model/__init__.py
```diff
@@ -180,6 +180,103 @@ def bulk_datum(self, doc):
...
+        name, output_doc : string, dict
+            The same name as what was passed in, and a doc that may be the same
+            instance as doc, a copy of doc, or a different dict altogether.
+        """
+        if name == 'start':
+            if self._start_doc is None:
+                self._start_doc = doc
+            else:
+                raise EventModelError(
+                    'SingleRunDocumentRouter received a second start document: uid {doc["uid"]}'
```
danielballan, line 218: Missing `f` from this f-string.

event_model/__init__.py
```diff
@@ -180,6 +180,103 @@ def bulk_datum(self, doc):
...
+
+        Returns
+        -------
+        descriptor document : dict
+        """
+        if 'descriptor' not in doc:
+            raise ValueError(f'document is not associated with a descriptor:\n{doc}')
+        elif doc['descriptor'] not in self._descriptors:
+            raise EventModelError(
+                'SingleRunDocumentRouter has not processed a descriptor with uid {doc["descriptor"]}'
```
danielballan, line 260: Missing `f`

event_model/tests/test_em.py
```diff
@@ -1116,7 +1189,7 @@ def all_factory(name, doc):
         return [header_collector], []
 
     rr = event_model.RunRouter([all_factory])
-    with pytest.warns(UserWarning, match='1.14.0'):
+    with pytest.warns(UserWarning, match='1.14.0'), pytest.raises(LocalException3):
```
danielballan, line 1192: I had no idea you could chain contexts like this!

event_model/__init__.py
jklynch: Nope.

event_model/__init__.py
jklynch: I'll be honest, I just used a couple different exceptions to see which ones would stick.

event_model/tests/test_em.py
jklynch: It works with files so I took a leap.

event_model/__init__.py
danielballan: A guideline which I picked up from a StackOverflow a long time ago is, "If it's going to work most of the time, begging forgiveness is faster. If it's going to fail often, ask permission." I checked if this guideline is still valid and it looks like it is.

```py
In [1]: d = {'a': 1, 'b': 2, 'c': 3, 'd': 4}                                                                                                                                                  

In [2]: %%timeit 
   ...: try: 
   ...:     d['c'] 
   ...: except KeyError: 
   ...:     ... 
   ...:                                                                                                                                                                                       
43.8 ns ± 1.12 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: %%timeit 
   ...: if 'c' in d: 
   ...:     d['c'] 
   ...:                                                                                                                                                                                       
68.3 ns ± 1.32 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Now, the _fail_ path is much slower...

```py
In [10]: %%timeit 
    ...: if 'x' in d: 
    ...:     d['c'] 
    ...:                                                                                                                                                                                      
39 ns ± 0.989 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [11]: %%timeit 
    ...: try: 
    ...:     d['x'] 
    ...: except KeyError: 
    ...:     ... 
    ...:                                                                                                                                                                                      
252 ns ± 2.95 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

...but in that path the slowness is unimportant because we are raising an unrecoverable exception.

If "Nope" means, "This micro-optimization is not worth bothering with," fair enough. If "Nope" means, "I get different results," then I'm intrigued even if this is just an academic exercise. :- )

event_model/__init__.py
jklynch: failure of testing hygiene

event_model/__init__.py
jklynch: failure of testing hygiene

event_model/__init__.py
jklynch: Sorry, I have been regretting this "Nope" for some time. I should have written up exactly what you have here with my reasoning which is that the work being done with the documents will be significantly more expensive than these checks and so no performance improvement would be realized at the cost of using the wrong language feature. I first saw this optimization in a "biocomputation with python" book and I (still) think if a program is slow because a hot loop is checking for a key in a dictionary the solution is not to replace `in` with `try/catch` but do something smarter than check for a key in the loop.

Apologies for my snippy response and the time you spent writing what I should have written.

danielballan: No worries at all, @jklynch.

Issue #170, opened 2020-04-29
subfactory called with descriptor document rather than start document


jklynch:
RunRouter erroneously delivers a descriptor document to subfactories rather than the start document. See [here](https://github.com/bluesky/event-model/blob/67b901535c81972ca9831bd1e653e964fea54399/event_model/__init__.py#L1206).

tacaswell:
I think this was closed b= #171 ?

PR #171, opened 2020-04-30
Fix RunRouter subfactory bug 

jklynch:
This PR addresses #170.

## Description
Added field `self._start_to_start_doc` to `RunRouter`. In `RunRouter.start(...)` store the start document by uid. In `RunRouter.descriptor(...)` retrieve the associated start document and give it to subfactory callbacks. In `RunRouter.stop(...)` remove the start document from `self._start_to_start_doc`. 

## How Has This Been Tested?
The test function `test_subfactory()` in `test_run_router.py` was written to target the bug reported in #170.

The test function `test_run_router()` has been moved from `test_em.py` to `test_run_router.py` to keep `RunRouter` tests together.


jklynch:
The initial commit adds a test to demonstrate the bug, but the test may not be quite right yet. At least it fails!

danielballan:
An elegant test.

I agree we are overdue for splitting up the event-model tests across more than more file. Would you mind moving [this embarrassingly monolithic "unit test"](https://github.com/bluesky/event-model/blob/67b901535c81972ca9831bd1e653e964fea54399/event_model/tests/test_em.py#L1097-L1284) into the new file as part of this PR, so we can keep the RunRouter tests together?

danielballan:
I confirmed that running the test in its final (fixed) form against the `master` branch version of event-model, I can reproduce the expected bug:

```
>       assert subfactory_documents["start"] == [run_bundle.start_doc]
E       AssertionError: assert [{'configurat...rimary', ...}] == [{'time': 158...742a544daa0'}]
```

Issue #3, opened 2016-04-01
need to provide a mechanism for versioning


tacaswell:
Need to be able to ask for 'latest' or some specific version of the schema.

danielballan:
Seems reasonable and straightforward to do. For prioritization purposes, have we encountered a need for this yet, or is it just a generally good idea?

tacaswell:
Just generally a good idea.

I suspect some of the bulk* work is going to point out flaws in the current schema.

PR #173, opened 2020-05-29
WIP: make the timestamps part of the payload optional

tacaswell:
## Description

Makes 'timestamps' option in the event schema, event page schema, and compose_* APIs.

## Motivation and Context

As discussed on today pilot call, the `'timestamps'` entry makes a lot of sense when the ultimate source of the data is a control systems (and fundamentally asynchronous) but makes much less sense for the output of data analysis.  

This is the minimal change to event model to allow this.  There will need to be a companion change to databroker to cope with data that may not have timestamps.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1506,7 +1506,7 @@ def compose_stop(*, start, event_counter, poison_pill,
     return doc
 
 
-def compose_event_page(*, descriptor, event_counter, data, timestamps, seq_num,
+def compose_event_page(*, descriptor, event_counter, data, timestamps=None, seq_num,
                        filled=None, uid=None, time=None, validate=True):
     N = len(seq_num)
     if uid is None:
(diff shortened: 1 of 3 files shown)


dylanmcreynolds:
Have you considered adding a field to schemas (and their documents) for "schema_version" so that people from the future can easily track which schema a particular instance was written against? This is a nice recommendation from the mongo folks (https://www.mongodb.com/blog/post/building-with-patterns-the-schema-versioning-pattern). It seems to me that this same logic would apply to non-mongo serializations as well. 

Granted, this particular schema change is easy to imply..."oh, there's not a datastamp? must have been after that was optional." Making it explicit might still be nice.

danielballan:
Related to https://github.com/bluesky/event-model/issues/3

Thanks for the Mongo reference, @dylanmcreynolds.

PR #174, opened 2020-05-29
DOC: expand documenation a bit more

tacaswell:
## Description

This makes some changes to the API docs to organize them a bit more and examples of how to decide where data should go in the model.


docs/source/index.rst
```diff
@@ -3,20 +3,44 @@
...
+
+For the full details and schema please see the :ref:`data_model` section.  This is a very quick guide to where
+you should look for / put different kinds of information
+
+* Information about your sample that you know before the measurement → *Start* Document
+* What experiment you intended to do → *Start* Document
+* Who you are / where you are → *Start* Document
+* References to external databases → *Start* Document
+* The Data™  → *Event* Document
+* Detector calibrations, dark frames, flat fields , or masks  → *Event* Document (probably in it's own stream)
```
danielballan, line 32:
```suggestion
* Detector calibrations, dark frames, flat fields , or masks  → *Event* Document (probably in its own stream)
```

tacaswell: I swear I don't do it to troll you.

Issue #145, opened 2020-02-18
Reinstate compatibility with jsonschema 2.x


danielballan:
In #79 we made bluesky compatible with the new jsonschema 3.x API. We also chose to _require_ that API and become incompatible with 2.x.

Unfortuantely, pip's dependency solver (unlike conda's) does not guarantee satisfying all the dependencies across steps; it just warns when it is going to ignore a pin. It often happens that our `jsonschema >=3` pin gets ignored.

It would be simple enough to do a version check on `jsonschema` at import time and use old-style validation code for jsonschema 2.x. It adds some complexity to our codebase, but given that this issue has come up a couple times and can be puzzling if you don't know what the problem is, I think it's worth considering.

PR #179, opened 2020-07-03
Add projections schema to start doc

dylanmcreynolds:
## Description
This is a take on another PR related to "techniques", which turned into "projections". https://github.com/bluesky/event-model/pull/130. I considered trying to commit to that PR, but this seemed cleaner, even though 130 has a lot of critical discussion in it. This really isn't very different from that.

## Motivation and Context
I won't got into too much detail here, but the basic thought was to address the need for mapping from a beamline's ontology to one ore more ontologies. I had in mind a case where a beamline is doing a scattering experiment and wants to provide the map to downstream users/application about how to map fields in NXsas fields. with that in mind, I had a few things in mind in the design:

- A Projection is a mapping to a vocabulary or ontologies. 
- The Projections object is collection of potential projections. No inherent limit to how many a beamline decides to project to.
- Each Projection is an object with a unique name and a value with fields that map to event_model objects.

I took no position on the names of Projections. For this all to work in, say, the NeXuS case, there will still need to be some sort of agreed format for specifying the fields to do this mapping. But I felt that that "unique string" was all we need for this  (uniqueness being enforced by being property names in the Projection object.

**EDIT:** I also updated compose_run to allow for projections to be written to the root of a run_start doc.

## How Has This Been Tested?
Added new test_projecitonpy including at least one failing case (which exposed a nice bug in my initial schema attempt).

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1625,7 +1625,7 @@ def compose_descriptor(*, start, streams, event_counter,
         partial(compose_event_page, descriptor=doc, event_counter=event_counter))
 
 
-def compose_run(*, uid=None, time=None, metadata=None, validate=True, projections=None):
+def compose_run(*, uid=None, time=None, metadata=None, validate=True):
     """
     Compose a RunStart document and factory functions for related documents.
 
@@ -1641,8 +1641,6 @@ def compose_run(*, uid=None, time=None, metadata=None, validate=True, projection
         Additional metadata include the document
     validate : boolean, optional
         Validate this document conforms to the schema.
-    projections: dict, optional
-        Additional projection data 
 
     Returns
     -------
@@ -1654,8 +1652,6 @@ def compose_run(*, uid=None, time=None, metadata=None, validate=True, projection
         time = ttime.time()
     if metadata is None:
         metadata = {}
-    if projections is not None:
-        metadata['projections'] = projections
     doc = dict(uid=uid, time=time, **metadata)
     # Define some mutable state to be shared internally by the closures composed
     # below.
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1625,7 +1625,7 @@ def compose_descriptor(*, start, streams, event_counter,
         partial(compose_event_page, descriptor=doc, event_counter=event_counter))
 
 
-def compose_run(*, uid=None, time=None, metadata=None, validate=True):
+def compose_run(*, uid=None, time=None, metadata=None, validate=True, projections=None):
     """
     Compose a RunStart document and factory functions for related documents.
 
@@ -1641,6 +1641,8 @@ def compose_run(*, uid=None, time=None, metadata=None, validate=True):
         Additional metadata include the document
     validate : boolean, optional
         Validate this document conforms to the schema.
+    projections: dict, optional
+        Additional projection data 
 
     Returns
     -------
@@ -1652,6 +1654,8 @@ def compose_run(*, uid=None, time=None, metadata=None, validate=True):
         time = ttime.time()
     if metadata is None:
         metadata = {}
+    if projections is not None:
+        metadata['projections'] = projections
     doc = dict(uid=uid, time=time, **metadata)
     # Define some mutable state to be shared internally by the closures composed
     # below.
--- a/event_model/tests/test_em.py
+++ b/event_model/tests/test_em.py
@@ -5,7 +5,6 @@ import pickle
 import event_model
 import numpy
 import pytest
-from jsonschema.exceptions import ValidationError
 
 def test_documents():
     dn = event_model.DocumentNames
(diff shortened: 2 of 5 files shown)


dylanmcreynolds:
@tacaswell and @danielballan, I debated adding to #130 vs. a whole new PR, especially since what we came up with was so similar. But if it was better in Tom's PR, then I suppose it's just one click away.

event_model/tests/test_em.py
```diff
@@ -1196,3 +1196,47 @@ def test_pickle_filler():
...
+    {
+        "uid": "abc",
+        "time": 0,
+        "projections": [
+            {
+                "name": "test",
+                "version": "42.0.0",
+                "configuration": {},
+                "projection": {
+                    'Entry/instrument/detector/data': {
```
tacaswell, line 1211: I am not sure we want to inherit the "full path" from nexus, but should be looking at what the keys in the `NXData` groups are. 

My understanding is that the `"insturment/..."` paths in nexus are directly analogous to the bodies of these projections and are still instrument / beam line / facility specific where as we want to map to the NXData definitions which re-map the devices specific mappings to a technique specific mapping.

dylanmcreynolds: Right. But I was assuming that the job of this schema was to be able to project from places in the document model (stream/field/etc) to a uniquely-identified spot in a vocabulary, and though NeXuS is what is probably desired, we needed to be open.  My example in the test was a bit of a trial balloon on how the mapping might look for NeXuS,  but once projections are adopted in event_model, there will still have to be some sort of coordination about string-to-NeXuS field conventions. With NeXuS being XML-ish, I happened to slip into an XPath-ish syntax.

If we want to be more explicit about NeXuS pathing here, that would be more limiting but save some effort in coordination.

prjemian, line 1211: `/entry/instrument/detector/data`

dylanmcreynolds: @prjemian ....does this reflect a consensus about how serialize pointers into fields in a NeXuS document? Are there other documented rules?

prjemian: Not sure I know how to answer that.  So, my answer has become verbose.

First off, the key of the projection dictionary is, I assume, the HDF5 address of the specific item in the **NeXus** file.  The NeXus [naming convention](https://manual.nexusformat.org/datarules.html#naming-conventions) is to use lower case.  I prepended a `/` since this should be an absolute address in the file.  Following NeXus convention for picking default names of groups [1], the path to a dataset `/entry/instrument/detector/data` names the `data` dataset of the NXdetector group (child of the NXinstrument group which is child of the NXentry group which is child of the file root).

[1] [default group names in NeXus](https://manual.nexusformat.org/rules.html#rules-for-structuring-information-in-nexus-files): For a hypothetical base class of `NXzzzz`, the default group would be named `zzzz` or `zzzz_1`, `zzzz_2` if there are several such.  In some NeXus classes, a specific name is required.  Mostly in the base classes, the name is more flexible.  Use the default naming unless you have a good reason to choose otherwise.

An [example is shown](https://apstools.readthedocs.io/en/latest/source/_filewriters.html#full-structure) in the documentation for a [NeXus filewriter callback](https://apstools.readthedocs.io/en/latest/source/_filewriters.html#hdf5-nexus-file-writers) I've written for the APS.  The example shows the raw data stored under `/entry/instrument/bluesky`. NeXus-style links position this data at various places in the NeXus structure satisfying the expectations of the NeXus base class structure.

prjemian: So our projection keys would point into the conventional NeXus base class structures.  A good test would be:

1. ingest from a NeXus file into databroker with declared projections
1. `nxwriter = apstools.filewriters.NXWriter()`
1. `for key, doc in db[-1]: nxwriter.receiver(key, doc)`
1. compare the new NeXus file with the original

jklynch:
The `test_em.py` file is _very_ long. I think projections tests should get their very own test file `test_projections.py`.

dylanmcreynolds:
I changed compose_run to actually allow a projection to be written to the run_start document.

@jklynch While adding a test, moved stuff out of test_em.py and into new test_projections.py

event_model/__init__.py
```diff
@@ -1625,7 +1625,7 @@ def compose_descriptor(*, start, streams, event_counter,
         partial(compose_event_page, descriptor=doc, event_counter=event_counter))
 
 
-def compose_run(*, uid=None, time=None, metadata=None, validate=True):
+def compose_run(*, uid=None, time=None, metadata=None, validate=True, projections=None):
```
danielballan, line 1628: All other non-required keys enter in through `metadata`. I think we should either add all of them to the signature or leave `projections` out of the signature.

dylanmcreynolds: I'll back out that change, let you add a projection to metadata.

dylanmcreynolds:
I added calculation fields. 

There is a corresponding PR for a projector in databroker that uses these projections: https://github.com/bluesky/databroker/pull/585

dylanmcreynolds:
Andi points out a concern that the the start document is already getting big at some beamlines, and that adding this to it might get in the way of searching.

danielballan:
@dylanmcreynolds Please add mention of this to the documentation, probably in this section, https://blueskyproject.io/event-model/data-model.html#run-start-document and note clearly that this is experimental.

Note: On today's pilot call, we agree to merge this (marked "experimental") by a week from today.

dylanmcreynolds:
> @dylanmcreynolds Please add mention of this to the documentation, probably in this section, https://blueskyproject.io/event-model/data-model.html#run-start-document and note clearly that this is experimental.
> 
> Note: On today's pilot call, we agree to merge this (marked "experimental") by a week from today.

Committed a brief note there.

danielballan:
We agreed to merge this in order to lower the barrier for people to play with it, on the understanding that we will very likely make backward-incompatible changes as we gain more experience with it.

Just waiting flake8 fixes, then good to go! :rocket:

dylanmcreynolds:
I bumped the version of jsonschema from 2 to 3 in the travis run that pegs jsonschema version and it runs. I don't know if this is appropriate or if it's better to remove that test run. @tacaswell , @danielballan ?

danielballan:
[Already covered this on Slack but re-summarizing for the public record:]

The build that pins `jsonschema==2` is meant to protect against the problem described in https://github.com/bluesky/event-model/issues/145, and it should stay until we need some feature in a newer version of jsonschema. I force-pushed to remove the commit that altered the CI configuration. It turns out that `jsconschema==2` was finding a minor mistake in the schema that was for whatever reason ignored by jsonschema 3.x. Fixed in 2ca0aec.

Issue #156, opened 2020-03-04
Proposed Schema Changes


danielballan:
**This is a long-term proposal, unrelated to databroker 1.0 or any of the upcoming release.**

The following changes have been previously proposed and discussed at various times. Many are mutually un-coupled and could be considered separately. At some point we should decide which ones we want to do and execute them all in one step, tagging event-model 2.0.0.

## Datum

1. Add a `time` key.
2. Add an `index` key with a unique monotonically increasing integer.
3. If (2) is accepted, remove `datum_id` which would no longer be needed because `(resource_uid, index)` would be a unique key. Event documents would still refer to a Datum via a construction like the current `datum_id`, i.e. a string like `{resource_uid}/{index}`
4. Remove some generality is favor of simplicity and efficiency: assume Datums are 1D slices. (All the ones we current have are, and it's hard to imagine a case that wouldn't be. Even if the external asset is "paragraphs in a Word document", you can slice on that.) Drop `datum_kwargs` and replace them with slice fields: `start`, `stop`, and `step`. All Datum documents would now have the same fields, and handlers could be simplified.

## Resource

5. Add `version`, referring the version of the `spec`, with an associated schema maintained with the handler. This will get a lot of use if (4) is accepted because all the handlers will be simplified.

## Event

6. Similar to (2), add an `index` key with a unique monotonically increasing integer.
7. Similar to (3), if (6) is accepted, remove `uid` which would no longer be needed because `(descriptor, index)` would be a unique key.

ambarb:
Is the Datum `time` key a timestamp per point (read)?  Or is it just the recorded time of a uid?  

If per point (or eventually maybe even a list/array of timestamps corresponding to a stack of images for a single point), then it would be good to have another key to describe how the time stamp was generated (default, cpu time recorded into the database).  However, timestamps can be generated by the firmware or IOC of the device or from a TTL pulse that doesn't come from the same IOC or something else.  For clarity and future proofing, it might be good to include timestamp origin/provenance key `time.orgin`

danielballan:
The proposed Datum `time` key is just when the document was generated (i.e. "the recorded time of a uid"), useful as internal metadata for grabbing batches of documents efficiently from the database or "replaying" document streams of old data with realistic timing.

tacaswell:
I 50/50 on adding the simpler spelling, but not convinced that we want to deprecate datum as-is.

danielballan:
[...]

PR #185, opened 2020-08-28
Include resource, datum in filler_state.

danielballan:
The `filler_state` is a threadsafe way to side-band information into
"coersion" functions that, for example, return the data as a dask array.
We added this so that the functions could have access to `shape` via the
descriptor. In some cases we also need access to the resource to do that
successfully. In the future, when we support variable-length chunks, we
may also want access to the datum.

Prototyped and interactively tested against FXI data in
https://gist.github.com/danielballan/b4fccac29cd77e7bdfd93eba775d9b81

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -905,6 +905,8 @@ class Filler(DocumentRouter):
                     resource_uid,
                     f"Datum with id {datum_id} refers to unknown Resource "
                     f"uid {resource_uid}") from err
+            self._current_state.resource = resource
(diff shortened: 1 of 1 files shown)


danielballan:
This whole feature feels like a giant hack, so I am slightly reluctant to extend it, but this PR constitutes a small, natural, and somewhat anticipated extension. When we rethink and simplify how bluesky handles external data (see https://github.com/bluesky/event-model/issues/156) this mechanism will likely not be needed, but in the short term is a very helpful escape hatch for making data readable.

PR #187, opened 2020-09-15
Discover handlers

gwbischof:
Moving discover_handlers from databroker to event_model.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -3,6 +3,7 @@ import collections.abc
 from distutils.version import LooseVersion
 import copy
 import json
+import entrypoints
 from enum import Enum
 from functools import partial
 import itertools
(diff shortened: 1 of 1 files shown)


danielballan:
Needs to declare the entrypoints dependency.

event_model/__init__.py
```diff
@@ -2211,3 +2212,47 @@ def default(self, obj):
                 return obj.item()
             return obj.tolist()
         return json.JSONEncoder.default(self, obj)
+
+def discover_handlers(entrypoint_group_name='databroker.handlers',
+                      skip_failures=True):
+    """
+    Discover handlers via entrypoints.
+    Parameters
```
danielballan, line 2220: Needs a blank line. If the original in databroker is missing one too that should be fixed there as well.

danielballan (commented):
This needs at least a smoke test and should be added to the docs.

PR #190, opened 2020-09-26
create use case doc

### LINKED ISSUE #189: Document design strategy behind  Tomography docstream
At the coalition call today, we discussed creating a documentation issue and PR to lay out the proposed design that we came up with for tiling / mosaic use cases in tomography. (FWIW, there's no documentation label available, so marked as enhancement.)

dylanmcreynolds:
Fixes #189 

A use case document, with a single use case about tomography tiling.


danielballan:
Power-cycled to run against the new GH Actions CI. Merge at will if CI passes.

danielballan (approved):

PR #192, opened 2020-10-06
Projection configuration fix

### LINKED ISSUE #183: Projection Schema could provide stronger validation
Several parts of the projection schema a mutually exclusive. For example, you can either have type=computed or a type=linked. Currently, the schema allows for linked types with computed attributes. It would be nice if the schema provided stronger validation for this case. There might be additional opportunities for stronger validation.

dylanmcreynolds:
Addresses an issue found that the projection schema does not have enough types.

## Description
The projection schema's "location" field contained "event" and "configuration". This PR adds "start". 

## Motivation and Context
The initial commits for projections and projectors contained the mistaken notion that the only two places in a docstream that might want to be projected are the start document and event fields. It was realized later that event descriptr configuration fields would also want to be project. This PR adds "start" to the projection location enum, and changes the meaning of "configuration" from the start document to the event descriptor.   Fixes

Additionally, fixes Fixes #183. This issue asked for more schema validation against calculated and liked fields. For example:

if type == calculated:
  calcuation field is also required
if type == linked:
  location field is also required

## How Has This Been Tested?
Tests in test_projections added for checking validation changes.

This PR will have  a corresponding PR in databroker.

### DIFF
--- a/event_model/schemas/run_start.json
+++ b/event_model/schemas/run_start.json
@@ -9,12 +9,13 @@
             "description": "Where to get the data from",
             "type": "object",
             "properties" : {
-                "type": {"type": "string", "enum": ["linked", "calculated", "static"], "description": "linked: a value linked from the data set, calculated: a value that requires calculation, static:  a value defined here in the projection "},
+                "type": {"enum": ["linked", "calculated", "static"], "description": "linked: a value linked from the data set, calculated: a value that requires calculation, static:  a value defined here in the projection "},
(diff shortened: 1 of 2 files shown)


dylanmcreynolds:
arghh....tests failing for jsonshema2

dylanmcreynolds:
Added skips in some tests for less than jsonschema3. h/t to @danielballan

danielballan (approved):

PR #195, opened 2020-10-30
Check for collisions in run router

danielballan:
While working on tests for https://github.com/bluesky/bluesky-live/pull/3 I accidentally fed `RunRouter` the same 'start' doc twice. It accepted it without complaint, leading to weird subsequent behavior that took me awhile to understand. This proactive collision-checking would have made the problem immediately clear.


jklynch (approved):
Great!

PR #196, opened 2020-11-16
Auth session field in start doc

dylanmcreynolds:
Adds auth_session field to start document schema

## Description
The optional auth_session field is intended to be a place to store filtering information that can be used by downstream systems when making decisions about authorization / privileges regarding the run.

## Motivation and Context
While this could be done by individual beamlines, it is desirable to make this part of the standard event model. The next step after this will be to go to the suitcase repos and index this field so that downstream processes can query it efficiently.

I made this a list of strings because the same run might have multiple known contexts. I had in mind "beamline A" and "proposal B" as distinct auth filtering contexts that could benefit.

## How Has This Been Tested?
New module test_auth.py includes tests for valid and invalid auth_sessions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,8 @@ target/
 #pycharm
 .idea/*
 
+#vscode
+.vscode/*
 
 #Ipython Notebook
 .ipynb_checkpoints
--- a/docs/source/data-model.rst
+++ b/docs/source/data-model.rst
@@ -115,7 +115,8 @@ A typical example
 .. code-block:: python
 
    # 'run start' document
-   {'detectors': ['random_walk:x'],
+   {'auth_session': ['bl42', 'proposal12345'],
+    'detectors': ['random_walk:x'],
     'hints': {'dimensions': [(['random_walk:dt'], 'primary')]},
     'motors': ('random_walk:dt',),
     'num_intervals': 2,
--- a/docs/source/data-model.rst
+++ b/docs/source/data-model.rst
@@ -115,7 +115,8 @@ A typical example
 .. code-block:: python
 
    # 'run start' document
-   {'auth_session': ['bl42', 'proposal12345'],
+   {'data_session': 'vist54321',
+    'data_groups': ['bl42', 'proposal12345'],
     'detectors': ['random_walk:x'],
     'hints': {'dimensions': [(['random_walk:dt'], 'primary')]},
     'motors': ('random_walk:dt',),
--- a/event_model/schemas/run_start.json
+++ b/event_model/schemas/run_start.json
@@ -76,7 +76,7 @@
         }
     },
     "properties": {
-        "auth_session": {
+        "data_session": {
             "type": "array", "items": {"type": "string"},              
             "description": "An optional filtering field to associate runs with authorization/privilege frameworks"
         },
--- a/event_model/schemas/run_start.json
+++ b/event_model/schemas/run_start.json
@@ -76,6 +76,10 @@
         }
     },
     "properties": {
+        "auth_session": {
+            "type": "array", "items": {"type": "string"},              
+            "description": "An optional filtering field to associate runs with authorization/privilege frameworks"
+        },
         "project": {
             "type": "string",
             "description": "Name of project that this run is part of"
--- a/event_model/schemas/run_start.json
+++ b/event_model/schemas/run_start.json
@@ -77,8 +77,12 @@
     },
     "properties": {
         "data_session": {
+            "type": "string",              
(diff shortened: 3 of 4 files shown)


stuartcampbell:
@dylanmcreynolds I thought we had decided to call this "Data Session"

dylanmcreynolds:
@stuartcampbell hmmm, I didn't take notes and it came up as auth_session in subsequent conversations. I'm game to make it whatever, though this is the right time, since I have it in four repos right now (two in splash, this one and suitcase-mongo.) You want I should change it?

danielballan:
For what it’s worth “data session” is what I remember us converging on too. There was something about maybe wanting “auth” for other uses, and this being particularly about who is authorized to see this data? But I have no strong views and will happily click the green button once a choice is made.

dylanmcreynolds:
OK, I'll make this change tomorrow:  data_session

tacaswell:
Is the order in the list meant to be meaningful?  It may be better to spell this as a dictionary that is `{'naming_authority': "token"}` so we do not have to search through a list going "does _this_ look like something I would have handed out to identify a data session?"

stuartcampbell:
The trouble is then, what is the key in the dictionary for the data session itself?  

If you want to specify a 'group' of people that are not associated with a facility,  beamline, proposal or (e)SAF what would go in this list or dictionary?

dylanmcreynolds:
@stuartcampbell In this PR, I am not proposing enforcing any governance on what can go in to the list or dict. If beamline wanted to give access to their cats, it would be on them to decide that "catz" is a valid group to add. Or am I seriously misunderstanding your question?

danielballan:
Summary of an impromptu conversation that sprung up over DAMA group coffee this afternoon:

Our (@stuartcampbell, @cryos, and me) recollection of the last meeting on this is that we would add a blessed key named `data_session` and map it to a single unique identifier (a string) that groups data that will have the same permissions. In typical practice, users would walk up to a beamline, declare a new "session", collect all their data, and then the session would close. The permissions associated with that session would be left up to an external system that can manage things like groups splitting or merging, being added or removed.

This gives us the ability to deal with cases like:
* one proposal that contains data with different access groups
* dealing with cross-facility data access where one facility's rules about who can see a certain data_session might be different than another's

The current implementation in this PR (a list of groups akin to UNIX groups) collapses these layers into one and makes those cases harder to deal with.

I would like to bring closure to this discussion promptly for your sake, in recognition that it is on your front burner. If you are content with a new key named `data_session` mapping to a unique identifier string, we're good. If we need more discussion, let's set a meeting soon and hash it out.

danielballan:
I posted a summary of a conversation with @stuartcampbell and @cryos but before I learned that, after that conversation, @dylanmcreynolds and @stuartcampbell came to an understanding. I have deleted my comment to avoid muddying the waters. I don't have strongly-held views or deep experience in this area. I'm just aiming to un-block progress. :+1:

stuartcampbell:
So, following up, @dylanmcreynolds and I had a chat this afternoon and cleared up some confusion that I had with thinking there was an implied metadata storage in the items in the list, but now I know its just a list of 'groups' that are allowed access to this data then I am happy.  So, I have no strong preference for a string or a list - and hence if @dylanmcreynolds has use cases for a list, then I am happy with that.

dylanmcreynolds:
I committed the change to "data_session" from "auth_session". As far as I know, this PR is good to go.

danielballan:
Forgive me for one last round of questions here. Changes to this schema are hard to back out of and I want to make sure we get this right on the first try.

When we chose the name "data session" I think we had in mind a UUID as the value. The UUID would have meaning to some external system that would map it to proposals / access groups / users.

This proposal writes the groups directly into the document, effectively removing a layer. Both approaches seem valid to me, and it may make sense to choose one or even both depending on the use case. Stroring a list of groups is more direct and simpler in the case where the documents remain under the management of a system that understands those groups.  Storing a UUID better enables the use case where documents may be accessed or moved between uncoordinated systems. For example, `proposal12345` may mean something different at NSLS-II than ALS.

I propose that we support both in the document model: an optional UUID for the session and an optional hard-coded list of groups that can access it. We could worry about the unique ID field in a future PR, but I bring it up now because I think the name "data session" fits better for that---for our original concept of `data_session: UUID`. This PR has evolved into something that I think we might better call "data groups" or "access groups".

dylanmcreynolds:
As discussed with @danielballan, I separated data_session (a string) and data_groups (list of strings).

danielballan (approved):

danielballan:
For the record: this is exactly in line with the discussion that took place during the DAMA group meeting on Monday involving @stuartcampbell, @cryos, @tacaswell, and others present, so I will take the liberty of merging it. Thanks for your patient attention to detail and group consensus here, @dylanmcreynolds.

PR #200, opened 2020-12-15
Actually pin jsonschema.

danielballan:
Follow up to #199


danielballan:
The output of `pip list` in the logs confirms that the builds for `jsonschema-version` set to `2` and `3` respectively installed

```
jsonschema                    2.6.0
```

and

```
jsonschema                    3.2.0
```

PR #201, opened 2020-12-15
Do not pin jsonschema in requirements.

### LINKED ISSUE #193: Unpin jsonschema in requirements?
The travis configuration effectively pins it to jsonschema 2.x because of an issue with PIP sometimes pulling down 2 instead of 3.

I raised this an issue in slack, and @tacaswell said "we should probably un-pin the jsonschema version".

danielballan:
Since we testing against jsonschema 2 and 3 in our CI, there is no need to pin to `>=3.0.0` in our requirements. Closes #193

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-jsonschema>=3.0.0
+jsonschema
 numpy


dylanmcreynolds (approved):
Makes sense to me.

PR #203, opened 2020-12-17
Release notes for 1.17.0

danielballan:


docs/source/release-history.rst
```diff
@@ -14,9 +28,6 @@ Added
 * The method :class:`~event_model.RunRouter.start` was enhanced to check for repeated runs with
   colliding uids, raising an ValueError when this occurs.
 
-Changed
```
danielballan, line 17: It seems we left an empty section here. Just cleaning this up while I'm here.

PR #210, opened 2021-05-11
record descriptor uuid before executing subfactory callbacks in RunRouter.descriptor()

### LINKED ISSUE #209: Store document uuids before executing callbacks in RunRouter
When method RunRouter.descriptor executes a callback which raises an exception, the RunRouter._descriptor_to_start and RunRouter._start_to_descriptors dictionaries are not updated starting on this [line](https://github.com/bluesky/event-model/blob/9d3651ef8a4d8e2c3f046524c92bd5ff0ceefa2d/event_model/__init__.py#L1351). This can cause a dictionary lookup failure at a later time when, for example, an event document is processed which references the offending descriptor document.

## Expected Behavior
The offending descriptor should be used to update RunRouter._start_to_descriptors and RunRouter._descriptor_to_start.

## Current Behavior
The dictionaries are not updated if a callback raises an exception.

## Possible Solution
Update the dictionaries before executing callbacks.

jklynch:
## Description
This PR moves the code in method `RunRouter.descriptor()` that records descriptor uuids in `RunRouter._start_to_descriptors` and `RunRouter._descriptor_to_start` so it will execute before
subfactory callbacks are executed. The intention  is to guarantee the dictionaries will be updated even if
a callback raises an exception.

## Motivation and Context
Currently if a callback executed by `RunRouter.descriptor()` raises an exception these
dictionaries will not be updated with the descriptor uuid. Subsequent calls to `RunRouter.event()`
with event documents referencing the same descriptor will then fail because the descriptor uuid
will not be found in `RunRouter._descriptor_to_start`.

Closes #209.

## How Has This Been Tested?
The test function `test_subfactory_callback_exception` has been added to `test_run_router.py`.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1317,19 +1317,27 @@ class RunRouter(DocumentRouter):
             self._subfactories[uid].extend(subfactories)
 
     def descriptor(self, descriptor_doc):
-        uid = descriptor_doc['uid']
+        descriptor_uid = descriptor_doc['uid']
         start_uid = descriptor_doc['run_start']
+
+        # Keep track of the RunStart UID -> [EventDescriptor UIDs] mapping for
+        # purposes of cleanup in stop().
+        self._start_to_descriptors[start_uid].append(descriptor_uid)
+        # Keep track of the EventDescriptor UID -> RunStartUID for filling
+        # purposes.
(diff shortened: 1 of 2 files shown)


danielballan:
Bump @gwbischof Can you review this one?

mrakitin (approved):
Looks good to me.

PR #211, opened 2021-05-18
ENH: scrub and offset data

tacaswell:
This is primarily useful if you want to obscure the origin of some data.

### DIFF
--- a/event_model/scrub.py
+++ b/event_model/scrub.py
@@ -1,0 +1,101 @@
+import event_model
+import itertools
+
+
+def scrub(hdrs, new_epoch, start_scan_id, filter_start=lambda x: x):
+    """
+    Scrubs a set of headers of most identifying information.
+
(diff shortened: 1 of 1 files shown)


tacaswell:
Example usage.

```python

from databroker._drivers.msgpack import BlueskyMsgpackCatalog
from suitcase.msgpack import Serializer

def sanitize_start(doc):
    doc["beamline_id"] = "MAD"
    for k in ("group", "project"):
        doc.pop(k, None)

    doc["sample"]["type"] = "Cool Material"
    return doc

cat = BlueskyMsgpackCatalog('/tmp/some_source/*.msgpack')

target = Serializer('/tmp/targetl')

for name, doc in scrub([cat[i] for i in cat], 0, 1000, sanitize_start):
    target(name, doc)

test = BlueskyMsgpackCatalog("/tmp/*msgpack")

```

PR #222, opened 2022-03-31
Add a new document type for "bulk" external resources with no predetermined shape.

### LINKED ISSUE #219: Add new document for "bulk" external resources
Currently the implementation of external data is limited to filling in data on a per-event per-field basis, basically the smallest quanta of data that the system can express.  While this works extremely well for something like image stacks of equal length, this works very poorly in cases where the data is natively ragged (single-photon detectors) or when there are many many relatively small data sets (e.g. scanned florescence data).  The difficulties vary from fully breaking the ability to put the data in an xarray to pathological performance problems.

One way out of this is to expand what can be "external".

## option A:

Expand the notion of "external" to be able to replace all of the rows in a field in a stream at once.   This would help significantly in cases like scanned florescence (because we can read out all of the data en-masse), but it would make it much much harder to pull out a _single_ events worth of data (but not impossible).   You could imagine saying "these col_datum is for index N -> N+m" so you get chunk level access.  A wrinkle that this is going to have is that we have the ability to emit multiple events with the same event number and we will silently ignore all but the last one (so if you re-take points it "does the right thing" 😱 .  This probably was a poor choice,but it made sense at the time).

The splicing to merge these into tiled is probably do-able.  We already do the splicing on a per-row basis and tiled already has a deep sense of chunking.  You would still keep the rule that the `col_datum` would have to go out before any event that refers to it goes out would still hold and I guess we would just duplicate the same `col_datum_uid` in all of the events.

A pro of this plan is that we can still use the structure of the `Events` to do the association between the "big" data and the "small" data and collate between multiple data sources in a very structural way.   That is, an `Event` that says "I have a motor position, an image, and a spectrum that are all 'at the same time for science'" is is much more powerful statement than "there are 3 streams in this Run (good luck guessing the names!) that _should_ be the same length and you can zip them...good luck!" in the same way that saying "this is a data frame" is a more powerful statement than "this is a dict of numpy arrays that I _promise_ are all the same length".

## option B:

The same as option A, but also replace multiple fields.  Again, this would help significantly with multi-channel spectral data as we could read up the whole x, y, channel, E hyper cube up at once.  You could still get the nice chunking, as option A but maybe even better.   This has all of the complexity of the above, but now spread across columns.   Also has the same pros.

At some point, this turns into an implementation of panda's block manager.

## option C:

Scale the notion up to be "the entire stream".

This gets around all of the splicing related issues (because it is everything).  This will handle cases like where we have some detector/device that outputs table like data in large volumes.  If we had an analogy of `stream_reosource` and maybe `stream_datum` (you could say "it is the whole stream, you'll only ever have one datum so stuff everything in the resource", but I think this is wrong as you may have a detector that writes an un-known number of fixed sized files.  It is better to positively say "and I expect another file" rather than to say "just glob all of those" because you will be able to detect when a file is missing (the document claimed it was there but you can not find it!) and resilient against the detector writing more later with the same file-prefix that is unrelated). 

This would also be very useful for single-photon detectors (where you do not know the size of the output until (well) after the fact) and complicated flyscans that may be already writing data out to (multiple) files.

A down side of this approach is that we would lose the ability to associate the rows with an event an another stream via indexing (but other methods like time stamping or including an "event index column" would still work.   The later may be the best approach to dealing with "ragged" data anyway.

## option D

Option C but provide a bunch of streams.

The down side of this is that it is a bunch of extra book keeping / semantics, but an upside would be that complicated files can provide mulitple streams and we will be able to open them only once.

-----

I think we should implement option D and maybe B, but do D first.   B has the possibility of providing some big speed improvements, but I think is actually much more complex to implement.  On the other had D will provide a very good escape hatch for some problems that have come up at both BNL and as DLS is thinking about adapting / wrapping / extending malcom.

Going down the path of D we should do our best to keep "feature" parity with the current scheme so we should still get a descriptor per out-sourced stream and we should make a best-effort to include event counts in the stop document.   To make this work we would have to add two new documents `stream_resource`, `stream_datum`.  I propose the following schema:

```yaml
- stream_resource:
   - spec : str # mapping to how to read this
     root : str # the "root" part of the resource path
     resource_path : str # the rest of the resource path, may want to call this 'uri'?
     path_semantics : {unix | windows}  # \ vs / maybe, maybe not? 
     uid : str # a uuid
     run_start : str  # mapping to the run start this belongs to
     stream_names : List[str]   # list of the stream names this provides
     resource_kwargs : Dict[str, Any]  # anything to pass into the handler
- stream_datum
   - resource_id : str refernce back to the stream_resource
     uid : a uuid
     stream_name : str # the same of the stream this is providing a block of
     block_id : int   # the order in the stream of this block
     datum_kwargs : Dict[str, Any]  # what to pass to call the handler with
     event_offset : int # what is the sequence number of the first event in this block
```

and signatures:

```python
class StreamHandler:
    def __init__(self, uri_path: str, **resource_kwargs):
        ...

    def get_block(self, stream_name: str, block_id: int, **datum_kwargs):
        ...

```

The order rules would be descriptor must come before a stream_resource, stream_resource must come before a stream_datum.  I think we should prohibit emitting more than one descriptor if using external streams (because otherwise spelling at what point to change descriptors gets too hard).   You can have as many stream_resource describing a stream as you want and each stream_resource can be referred to by as many stream_datum as you want, but `block_id` must be a unique run of integers with no gaps and be emitted in order.    event_offset must be unique and increasing, but can (obviously) have gaps.

maffettone:
Develop a `stream_resource` that manages an unknown number of contiguous `stream_datum`, with the potential for multiple streams. This is especially relevant when the data is expected to be ragged or has no pre-determined shape (number of rows). 

The potential for multiple streams makes this construction slightly different from `resource` in that multiple counters, and stream names need to be tracked within a closure. 

## Description
- Addition of 2 new document types and json specifications 
    - stream_datum
    - stream_resource
- stream_datum_page was not included, as this felt like a hat on a hat. 
- Refactor `ComposeRunBundle` from a namedtuple into an iterable dataclass. This is for backwards compatibility and extensibility should new documents be needed in the future.
- Compose streaming run bundle as extension to previous return value. (Passes all previous tests)
- Extend DocumentRouter for new documents
- Extend RunRouter for new documents
- Extend Filler with methods for new documents
- Documentation Changes:
    - Add sections to data-model.rst, and update intro highlighting these components as (Experimental).

### TODO: 
- Documentation

## Motivation and Context
Currently the implementation of external data is limited to filling in data on a per-event per-field basis, basically the smallest quanta of data that the system can express. While this works extremely well for something like image stacks of equal length, this works very poorly in cases where the data is natively ragged (single-photon detectors) or when there are many many relatively small data sets (e.g. scanned florescence data). The difficulties vary from fully breaking the ability to put the data in an xarray to pathological performance problems.

Closes #219.

## How Has This Been Tested?

Testing environment included all `event_model` and `bluesky` requirements in Python 3.9. 

All previous tests pass. 

Tests added:
- `test_em.test_compose_stream_resource`
- `test_em.test_document_router_streams_smoke_test`
- `test_run_router.test_run_router_streams`


maffettone:
Before going much deeper and embedding these new documents into the DocumentRouter, and Filler classes, I wanted to get a partial review (@tacaswell ) and clarify that my approach here was sensible. 

A given `stream_resource` can point to multiple `stream_datum` with unique names. I.e.  a single resource could point to the output stream of a single photon detector and the output stream of some thermocouple. Each datum can contain one or many rows (frames of the detector or voltage/temperature pairs of the thermocouple), but a single block_id, for the sake of constructing the full column in retrospect. The event_offset is there for the sake of book-keeping in that block 0 and block 1 may actually be separated by a gap?

event_model/schemas/stream_datum.json
```diff
@@ -0,0 +1,40 @@
...
+            "description": "The name of the stream that this Datum is providing a block of."
+        },
+        "block_id": {
+            "type": "integer",
+            "description": "The order in the stream of this block of data. This must be contiguous for a given stream."
+        },
+        "event_offset": {
+            "type": "integer",
+            "description": "The sequence number of the first event in this block. This increasing value allows the presence of gaps."
+        }
```
danielballan, line 26: We need some concept of `event_count` to tell us how many Events' worth of data this is going to give us.

maffettone: Addressed initially in https://github.com/bluesky/event-model/pull/222#:~:text=count%20to%20stream-,9c6b8da,-Merge%20state

event_model/__init__.py
```diff
@@ -1563,13 +1629,29 @@ def _is_array(checker, instance):
...
+class ComposeRunBundle:
+    """Extensible compose run bundle. This maintains backward compatibility by unpacking into a basic
+    run bundle (start, compose_descriptor, compose_resource, stop). Further extensions are optional and
+    require keyword referencing (i.e. compose_stream_resource).
+    """
+    start_doc: dict
+    compose_descriptor: callable
+    compose_resource: callable
+    compose_stop: callable
+    compose_stream_resource: Union[callable, None] = None
```
tacaswell, line 1643:
```suggestion
    compose_stream_resource: Optional[callable] = None
```

tiny nit, but I think it reads better and is the suggested spelling in this use case.

event_model/__init__.py
```diff
@@ -1,5 +1,7 @@
 from collections import defaultdict, deque, namedtuple
 import collections.abc
+from dataclasses import dataclass
+from typing import Union
```
tacaswell, line 4:
```suggestion
from typing import Optional
```

event_model/schemas/stream_datum.json
```diff
@@ -0,0 +1,44 @@
...
+
+    },
+    "required": [
+        "datum_kwargs",
+        "stream_resource",
+        "uid",
+        "stream_name",
+        "block_id"
```
tacaswell, line 38:
```suggestion
        "block_id",
        "event_count"
```

event_model/schemas/stream_datum.json
tacaswell: If we allow each datum to have a variable number of events in it (which is something we need for many of the flying use cases, it is much simpler to bundle and emit "what ever I just got in the last second" than "bundle and emit when I have _exactly_ 50 events" and we will in general have less events in the final datum), than if the user says "I want events `[10_000:15_000]`" then if all we have is block index then we have to iterative count from the beginning.  On the other hand if we have the offsets in each document we can say "dear database, please give me all datum with offsets in this range and the one with the highest block ID that is less that start" which should be much more efficient.

We ales need this when you are working in a context where you need to pre-allocate space for the data before you load it.

event_model/__init__.py
```diff
@@ -1620,6 +1702,54 @@ def compose_resource(*, spec, root, resource_path, resource_kwargs,
         partial(compose_datum_page, resource=doc, counter=counter))
 
 
+def compose_stream_datum(*, stream_resource, stream_name, counter, datum_kwargs,
+                         event_count=1, event_offset=0, validate=True):
+    resource_uid = stream_resource['uid']
```
tacaswell, line 1707: Should we add a check here that the resource document actually know about the stream we are trying to make?

On one hand YOLO why does the resource need to know what streams it will contain?  On the other hand, the resource document currently advertises the expected streams, those we should make sure we do not generate a stream of documents that is not internally consistent.

History has shown that if we have a "MUST" (or "SHOULD") requirement that is not enforced by code, we will end up with reams of documents that are incorrect.  The primary example here is all of the issues we have been cleaning up with the shapes!

maffettone: Adjustments made in source and testing for validation in https://github.com/bluesky/event-model/pull/222/commits/9082b82c92e0c7833555dadb12c3efaf17b478fe

event_model/schemas/stream_datum.json
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+    },
+    "required": [
+        "datum_kwargs",
+        "stream_resource",
+        "uid",
+        "stream_name",
+        "block_idx",
+        "event_count"
```
tacaswell, line 39:
```suggestion
        "event_count",
        "event_offset"
```

coretl:
We're happy with this so happy to see it merged whenever

danielballan:
@tacaswell Is this ready to go? I will run the release process when this is merged.

maffettone:
Unsure why I left Documentation under the TODO's list. Perhaps just because it was failing the documentation Action.

tacaswell:
@coretl Lets see what breaks when you use this in anger 😈 .

PR #223, opened 2022-05-24
reserve time and seq_num keys

gwbischof:
user-defined data_keys called "time" or "seq_num" conflict with with bluesky generated keys of the same name.

### DIFF
--- a/event_model/schemas/event_descriptor.json
+++ b/event_model/schemas/event_descriptor.json
@@ -97,7 +97,13 @@
                 },
                 "type": "object",
                 "description": "This describes the data in the Event Documents.",
-                "title": "data_keys"
+                "title": "data_keys",
+                "not": {
+                    "required": [
+                        "time",
+                        "seq_num"
+                    ]
+                }
             },
             "uid": {
                 "type": "string",


gwbischof:
This is a useful stackoverflow post about using `not` in jsonschema: https://stackoverflow.com/questions/61059683/how-to-conditionally-forbid-properties-based-on-presence-of-other-properties-in/61062869#61062869

tacaswell:
Do we have any users _currently_ using them?

dmgav:
@tacaswell https://github.com/NSLS-II-SRX/profile_collection/blob/06b6d22ba27f75ff1b775c794f68889a02278be3/startup/32-zebra.py#L847 If I understand correctly, it has the same meaning (and probably values) as `time` generated by Bluesky and could probably be removed.

danielballan:
The intent is to pair this with a migration tool (similar to "shape fixer") to alter the names on existing data.

danielballan:
This needs:

- [ ] a substantial rebase, as the system for encoding schemas has evolved substantially
- [ ] a clear story for how to deal with existing usage

PR #227, opened 2022-10-11
Use importlib.metadata to determine jsonschema version

callumforrester:
Use `importlib.metadata` to determine `jsonschema` version. `event-model` contains some conditional logic depending on the version.

## Description
Pure refactor: no change in behaviour, this just uses a more up-to-date method of determining the version.

## Motivation and Context
The current method is deprecated and produces the following warning:
```shell
DeprecationWarning: Accessing jsonschema.__version__ is deprecated and will be removed in a future release. Use importlib.metadata directly to query for jsonschema's version.
```

## How Has This Been Tested?
TODO

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -9,6 +9,7 @@ import itertools
 import inspect
 import os
 from pkg_resources import resource_filename as rs_fn
+from importlib.metadata import version
 import threading
 import time as ttime
 import types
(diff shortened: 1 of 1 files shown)


mrakitin (approved):
Thanks fore the fix, @callumforrester! The description and the fix are clear.

PR #233, opened 2022-12-09
Include importlib-metadata for 3.7.

### LINKED ISSUE #230: Python minimum version/using 3.8+ features
https://github.com/bluesky/event-model/blob/eac55ca21e54328a7fc1f16fea71e56460ed6a58/event_model/__init__.py#L14

`importlib.metadata` was added in python 3.8, but event-model does not either specify 3.8 as minimum version or include the [backport pypi package](https://pypi.org/project/importlib-metadata/) as a conditional dependency.

Thus it will install on python 3.7 but fail upon use unless the pypi backport is installed otherwise.

danielballan:
Closes #230

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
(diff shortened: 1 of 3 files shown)


mrakitin (approved):

AbbyGi (approved):
Looks good

PR #239, opened 2023-01-18
235 add typing support to event model

### LINKED ISSUE #235: Add typing support to event_model

evalott100:
Added Type hints to event_model/__init__.py.

## Description
Added type hints to the compose functions. Added type hints to almost all the other parts of the module too, save some places where types were a little more fussy.

Also added mypy to the continuous dev environment.

## Motivation and Context
Closes #235 

## How Has This Been Tested?
Tested by seeing if unit tests remain unchanged.  Tested types by just running mypy.


event_model/__init__.py
```diff
@@ -1641,109 +1818,176 @@ class ComposeRunBundle:
...
+)
+
+
+def compose_datum(
+    *,
+    resource: dict,
+    counter: Iterator,
+    datum_kwargs: Iterable,
+    validate: bool = True,
+) -> dict:
```
coretl, line 1856: Please can we tie down the return values from `compose_*` a bit more tightly, I'd like to see a `documents.py` that contains `TypedDict`s that are returned from these functions.

A bit like
https://github.com/bluesky/bluesky/blob/d2b6bddd4415cdb2946fe8d552bd704bc9932c11/bluesky/protocols.py#L13-L30

evalott100:
I've been working on generating the schema from TypeDicts for more robust typing, the generated schema are [here](https://github.com/evalott100/typed-dict-to-pydantic-basemodel/tree/main/typed_dict_to_pydantic_basemodel/schemas/generated_json_schema), generated from [these models](https://github.com/evalott100/typed-dict-to-pydantic-basemodel/tree/main/typed_dict_to_pydantic_basemodel/src/models/pydantic)

@coretl  and I agree it makes more sense to sort out #240 before trying to implement this though.

PR #242, opened 2023-02-17
Use packaging.version instead of importlib.metadata.version

callumforrester:
Use packaging.version instead of importlib.metadata.version

## Description
Change the function used to determine the package version in response to a deprecation warning.

## Motivation and Context
 The distutils package has been deprecated for Python>=3.10 and will be removed in 3.12 per https://peps.python.org/pep-0632/
Now seeing the following when using event-model:
```shell
Traceback (most recent call last):
  File "/home/vid18871/projects/dodal/src/dodal/util.py", line 7, in <module>
    from bluesky.protocols import (
  File "/home/vid18871/projects/dodal/venv/lib64/python3.8/site-packages/bluesky/__init__.py", line 6, in <module>
    from .run_engine import RunEngine  # noqa: F401
  File "/home/vid18871/projects/dodal/venv/lib64/python3.8/site-packages/bluesky/run_engine.py", line 16, in <module>
    from .bundlers import RunBundler, maybe_await
  File "/home/vid18871/projects/dodal/venv/lib64/python3.8/site-packages/bluesky/bundlers.py", line 6, in <module>
    from event_model import DocumentNames
  File "/home/vid18871/projects/dodal/venv/lib64/python3.8/site-packages/event_model/__init__.py", line 1609, in <module>
    if LooseVersion(version("jsonschema")) >= LooseVersion("3.0.0"):
  File "/home/vid18871/projects/dodal/venv/lib64/python3.8/site-packages/setuptools/_distutils/version.py", line 55, in __init__
    warnings.warn(
DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
```

## How Has This Been Tested?
TBC!
The docs build passes in CI, which validates the case for jsonschema>=3.0.0. I have tested locally with both the latest version and 2.6.0 for the other case.

PR #246, opened 2023-03-17
Update requirements.txt

ksunden:
This was introduced as a dependency by #242

It is already a direct dependency of (among others) mpl and ophyd, but in the Wright Group's minimal docker environment none of those packages which have this required already were actually installed.


ksunden:
Closing as fixed by #245

PR #247, opened 2023-03-20
235 add typing support to event model

### LINKED ISSUE #235: Add typing support to event_model

evalott100:
Closes #235 

Adds TypedDicts and code to generate schemas from TypedDicts. It's necessary to merge  #245 prior to this so black formatting doesn't cause problems.


event_model/documents/event_descriptor.py
```diff
@@ -0,0 +1,129 @@
...
+        str, Field(description="The source (ex piece of hardware) of the data.")
+    ]
+
+
+if TYPE_CHECKING:
+    ObjectHints = Any
+else:
+
+    class ObjectHints(TypedDict):
+        __root__: Annotated[Any, Field(title="Object Hints")]
```
coretl, line 52: As discussed in person, this should be `Annotated[Any, AsRef("ObjectHints")]` and the creation of a `BaseModel` class with `__root__` annotation can happen in the schema creation

event_model/documents/_type_wrapper.py
```diff
@@ -0,0 +1,33 @@
...
+    pydantic_installed = False
+
+    def Field(*args, **kwargs):
+        ...
+
+
+extra_schema = {}
+
+
+def add_extra_schema(schema):
```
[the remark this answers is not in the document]
evalott100: A call to `reveal_type(cls)` gives 
```
Runtime type is '_TypedDictMeta'
<class 'event_model.documents.run_start.RunStart'>
```

event_model/documents/event_descriptor.py
```diff
@@ -0,0 +1,121 @@
+from typing import Any, Dict, List
+
+from typing_extensions import Annotated, NotRequired, Literal, TypedDict
+
+from .generate.type_wrapper import Field, AsRef
+
+
+class DataKey(TypedDict):
+    external: NotRequired[
```
coretl, line 9: Can we have docstrings for these intermediate classes where they exist in the existing json schemas please?

event_model/documents/event_descriptor.py
```diff
@@ -0,0 +1,121 @@
...
+        Annotated[
+            str,
+            Field(
+                description="Where the data is stored if it is stored external to the events.",
+                regex=r"^[A-Z]+:?",
```
coretl, line 14: These don't seem to make it into the json schema...

event_model/documents/event_descriptor.py
evalott100: Bugs ironed out in the latest commit, now all sub-typed dicts (those referenced in the annotations of an attribute in a typeddict) are swapped out for generated basemodels.

coretl (commented):
That's it for now, will have another look after Easter...

event_model/schemas/datum.json
```diff
@@ -1,25 +1,27 @@
...
-            "type": "string",
-            "description": "Globally unique identifier for this Datum (akin to 'uid' for other Document types), typically formatted as '<resource>/<integer>'"
-        }
-    },
-    "required": [
-        "datum_kwargs",
-        "resource",
-        "datum_id"
-    ],
-    "additionalProperties": false,
```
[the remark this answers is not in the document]
evalott100: It's not there when I generate it now, not sure exactly why, probably to do with a slight change in the schema generation config

event_model/schemas/event_descriptor.json
```diff
@@ -1,150 +1,142 @@
...
+               "title": "Data Keys",
+               "description": "This describes the data stored alongside it in this configuration object.",
+               "type": "object",
+               "additionalProperties": {
+                  "$ref": "#/definitions/DataKey"
+               }
             }
-        },
-        "patternProperties": {
-            "^([^./]+)$": {"$ref": "#/definitions/data_type"}},
```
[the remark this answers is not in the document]
evalott100: From my [reading](https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties)  this will have any string with no `./` becoming a data_type. Am I right about this @danielballan

event_model/documents/generate/typeddict_to_schema.py
```diff
@@ -0,0 +1,530 @@
...
+
+
+def sort_jsonschema(schema: dict) -> dict:
+    """Sorts the schema properties keys alphabetically by key name, exchanging the
+    properties dicts for OrderedDicts"""
+    for key in schema:
+        if key == "properties":
+            schema[key] = OrderedDict(
+                sorted(list(schema[key].items()), key=lambda x: x[0])
+            )
```
coretl, line 431: Nit: As of python 3.6 all dictionaries keep insertion order, so technically you could get away with a standard dict, but this is probably more readable

event_model/__init__.py
```diff
@@ -1801,46 +1869,62 @@ def __iter__(self):
...
+    )
     if validate:
         schema_validators[DocumentNames.datum].validate(doc)
     return doc
 
 
-def compose_datum_page(*, resource, counter, datum_kwargs, validate=True):
+def compose_datum_page(
+    *,
+    resource: dict,
```
coretl, line 1892:
```suggestion
    resource: Resource,
```

event_model/__init__.py
```diff
@@ -1953,15 +2037,15 @@ def compose_stream_resource(
...
-    start,
-    event_counters,
-    poison_pill,
-    exit_status="success",
-    reason="",
-    uid=None,
-    time=None,
-    validate=True,
-):
+    start: dict,
```
coretl, line 2040:
```suggestion
    start: RunStart,
```
etc for the other compose functions

event_model/documents/generate/typeddict_to_schema.py
evalott100: Yep the alternative with insertions would still require a sort on the keys, I prefer this

event_model/documents/generate/typeddict_to_schema.py
```diff
@@ -421,21 +421,42 @@ def strip_newline_literal(schema: dict):
...
+
+
+def sort_schema(schema: dict) -> dict:
+    if isinstance(schema, dict):
+        schema = OrderedDict(
+            sorted(list(schema.items()), key=lambda x: SortOrder.get(x[0], 8))
+        )
+
+        for key in schema:
+            if key in ("defintions", "properties"):
```
coretl, line 439:
```suggestion
            if key in ("definitions", "properties"):
```

event_model/documents/run_start.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+RUN_START_EXTRA_SCHEMA = {
+    "definitions": {
+        "Projection": {
+            "allOf": [
+                {
+                    "if": {
+                        "allOf": [
+                            {"properties": {"location": {"enum": "configuration"}}},
```
coretl, line 88: These seems to be `... "enum": ["configuration"] ...` in the original doc, not sure if this is equivalent?

event_model/schemas/event_page.json
```diff
@@ -3,61 +3,83 @@
...
       },
       "filled": {
+         "title": "Filled",
          "description": "Mapping each of the keys of externally-stored data to an array containing the boolean False, indicating that the data has not been loaded, or to foreign keys (moved here from 'data' when the data was loaded)",
-         "$ref": "#/definitions/dataframe_for_filled"
+         "allOf": [
+            {
+               "$ref": "#/definitions/DataframeForFilled"
+            }
+         ]
```
coretl, line 54: Is there any way to make pydantic put in just `"#/definitions/dataframe_for_filled"` rather than `"allof"...`?

coretl (commented):
The schemas are looking pretty close now, thanks for reordering them. A few comments...

event_model/documents/run_start.py
evalott100: Must have been lost when it was brought over, I'll add it!

event_model/schemas/event_page.json
evalott100: @coretl 

Doesn't look as pretty but the behaviour is the same as if the "allOf" weren't there. This seems to be pydantic's behaviour whenever a basemodel is included as a type in another basemodel:

``` python
class X(TypedDict):
    x: Annotated[int, Field(description="x field")]

modelX = create_model(X.__name__, __config__=Config, **get_annotations(X))



class Y(TypedDict):
    x: Annotated[modelX, Field(description="x field")]
    y: Annotated[int, Field(description="y field")]

modelY = create_model(Y.__name__, __config__=Config, **get_annotations(Y))

```
gives us:

```
{'definitions': {'X': {'properties': {'x': {'description': 'x field',
                                            'title': 'X',
                                            'type': 'integer'}},
                       'required': ['x'],
                       'title': 'X',
                       'type': 'object'}},
 'properties': {'x': {'allOf': [{'$ref': '#/definitions/X'}],
                      'description': 'x field',
                      'title': 'X'},
                'y': {'description': 'y field',
                      'title': 'Y',
                      'type': 'integer'}},
 'required': ['x', 'y'],
 'title': 'Y',
 'type': 'object'}
```

You think it's worth writing another function, ran after generation, which replaces any

``` python
"allOf": [{"$ref": "#/definition/some_class"}]
```
with

``` python
"$ref": "#/definition/some_class"
```
?

coretl: If it is functionally equivalent let's leave it like it is. It might be fixed in Pydantic 2 anyway...

evalott100:
## Itemising the differences in the schema here
* Capitalised definition names

https://github.com/evalott100/event-model/blob/e91f648dc74c294a97ceac497f25cb6d6d193d87/event_model/schemas/run_start.json#L5-L14

coretl (approved):

danielballan (approved):

PR #257, opened 2023-04-20
Added event_counters as an optional kwarg to compose_run

### LINKED ISSUE #256: Allow event counters to be defined outside of a run
Currently a `compose_run` creates its own event_counter, which it iterates on itself. When we implement compose functions in the bluesky `RunBundler` for the `rewind()` and `_interruptions_counter` we will need a reference to these counters to pass around. We should therefore allow the kwarg `event_counter` in the `compose_run` function.

evalott100:
Closes #256 

Also renames `event_counter` to `event_counters` since it is a dict of individual event counters.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1954,7 +1954,7 @@ def compose_stream_resource(
 def compose_stop(
     *,
     start,
-    event_counter,
+    event_counters,
     poison_pill,
     exit_status="success",
     reason="",
@@ -1977,7 +1977,7 @@ def compose_stop(
         "run_start": start["uid"],
         "exit_status": exit_status,
         "reason": reason,
-        "num_events": {k: v - 1 for k, v in event_counter.items()},
+        "num_events": {k: v - 1 for k, v in event_counters.items()},
     }
     if validate:
         schema_validators[DocumentNames.stop].validate(doc)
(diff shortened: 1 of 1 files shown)


event_model/__init__.py
```diff
@@ -2165,21 +2171,23 @@ def compose_run(*, uid=None, time=None, metadata=None, validate=True):
     # Define some mutable state to be shared internally by the closures composed
     # below.
     streams = {}
```
coretl, line 2173:
```suggestion
    streams = {}
    if event_counters is None:
        event_counters = {}
```

coretl (changes requested):

event_model/__init__.py
evalott100: I prefer the or syntax for this, though it would require changing the others and I'm not sure which nsls prefer? @danielballan 

For now I'll use and if `is None` and we can change it another time if need be.

coretl (approved):
Looks fine to me

coretl:
@danielballan what code actually uses these compose functions? Bluesky doesn't, I guess some ophyd flyers might, but I can't think what else might call `compose_descriptor` for instance?

danielballan:
It is used in [bluesky-live](https://github.com/bluesky/bluesky-live/search?q=compose_run) which has been sitting on the back burner, to be visited once the dust settles on Tiled. It may also be used in some smaller projects related to automation.

danielballan:
I would be in favor of refactor the bluesky `RunBundler` to use these. This was always the intention, just has not happened yet.

danielballan (approved):

PR #260, opened 2023-05-05
Removed an off by one error

### LINKED ISSUE #259: Slight difference between the current bluesky bundler and event model
# First problem

In the below snippet of [code](https://github.com/bluesky/event-model/blob/86474f030294436d33be4ce6d64082a2fc0a8913/event_model/__init__.py#L2127-#L2129) , the descriptor name is added to the event_list as a key, with number of events of 1, even if the descriptor hasn't had any code inserted.

``` python
    if name not in streams:
        streams[name] = set(data_keys)
        event_counters[name] = 1
```

This causes a `num_events` to be inaccurate causing the following behaviour:

```   python
    assert h.stop['num_events'] == {'primary': 0}
```
instead of what is depicted in the following [assertion](https://github.com/bluesky/bluesky/blob/1032136434a318306940fa64af34354331bf78e3/bluesky/tests/test_run_engine.py#L1576), as well as

```   python
    assert h.stop['num_events'] == {'primary': 0, 'baseline': 2}
```

in this  [assertion](https://github.com/bluesky/bluesky/blob/1032136434a318306940fa64af34354331bf78e3/bluesky/tests/test_run_engine.py#L1595).

Do we keep the current bluesky behaviour by filtering out descriptors with an event_counter of 1, or do we add in the descriptor name with `num_events` of 0 directly? @danielballan  @coretl

# Second problem
The following [code](https://github.com/bluesky/bluesky/blob/fddf4aeeb387019e65aea879506dfcae34024b8e/bluesky/bundlers.py#L111-#L113) in the bluesky bundler isn't accounted for in event_model, it can be added very easily on `compose_stop`, let me know what you think.
 
``` python
            if bundle_name is None:
                # rare but possible via Msg('create', name='primary')
                continue
```

evalott100:
Closes #259 

Fixes the problems mentioned in the above issue by 

* Filtering out `None` key descriptor and descriptors with only one element from the `num_events` in the `close_run` doc.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1977,9 +1977,7 @@ def compose_stop(
         "run_start": start["uid"],
         "exit_status": exit_status,
         "reason": reason,
-        "num_events": {
-            k: v - 1 for k, v in event_counters.items() if k is not None and v != 1
-        },
+        "num_events": {k: v - 1 for k, v in event_counters.items() if k is not None},
     }
     if validate:
         schema_validators[DocumentNames.stop].validate(doc)
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1977,7 +1977,9 @@ def compose_stop(
         "run_start": start["uid"],
         "exit_status": exit_status,
         "reason": reason,
-        "num_events": {k: v - 1 for k, v in event_counters.items()},
+        "num_events": {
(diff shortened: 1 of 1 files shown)


evalott100:
Closing since the `name=None` in a `Msg` should throw an error elsewhere

PR #262, opened 2023-05-15
Switch from partials to classes

### LINKED ISSUE #261: Switch from `partials` to Compose classes
Currently, compose functions return documents (a dict), and partials of further compose functions, all inside a Bundle: 

https://github.com/bluesky/event-model/blob/86474f030294436d33be4ce6d64082a2fc0a8913/event_model/__init__.py#L2180-#L2196

Instead of namedtuple run bundles, and partials, we should use dataclasses, and compose classes of the form
``` python
class ComposeDatum:
    resource: Resource
    counter: int

    def __call__(self, uid:str) -> Datum:
        return Datum(uid=uid, resource=self.resource)
    
compose_datum = back_compat_wrapper(ComposeDatum)

def compose_datum(resource, counter, uid) -> Datum:
    return ComposeDatum(resource, counter)(uid)
```

evalott100:
Closes #261 

Switched from using partials to using dataclasses with a `__call__` method.
Defined `compose_*` functions as wrappers to these classes.


event_model/__init__.py
```diff
@@ -1797,353 +1872,647 @@ def __iter__(self):
...
-    "ComposeResourceBundle", "resource_doc compose_datum compose_datum_page"
-)
-ComposeStreamResourceBundle = namedtuple(
-    "ComposeStreamResourceBundle", "stream_resource_doc compose_stream_data"
-)
+@dataclass
+class ComposeResourceBundle:
+    resource_doc: Resource
+    compose_datum: Callable
+    compose_datum_page: Callable
```
coretl, line 1879:
```suggestion
    compose_datum: ComposeDatum
    compose_datum_page: ComposeDatumPage
```
Also, could you move this class to just above Compose Resource please?

event_model/__init__.py
```diff
@@ -1780,13 +1854,14 @@ class ComposeRunBundle:
...
 
-    start_doc: dict
-    compose_descriptor: callable
-    compose_resource: callable
-    compose_stop: callable
-    compose_stream_resource: Optional[callable] = None
+    start_doc: RunStart
+    compose_descriptor: Callable
+    compose_resource: Callable
+    compose_stop: Callable
```
coretl, line 1860:
```suggestion
    compose_descriptor: ComposeDescriptor
    compose_resource: ComposeResource
    compose_stop: ComposeStop
```
Also, could you move this class to just above ComposeRun please?

event_model/__init__.py
```diff
@@ -1797,353 +1872,647 @@ def __iter__(self):
...
-        schema_validators[DocumentNames.event_page].validate(doc)
-        if not (descriptor["data_keys"].keys() == data.keys() == timestamps.keys()):
-            raise EventModelValidationError(
-                "These sets of keys must match:\n"
-                "event['data'].keys(): {}\n"
-                "event['timestamps'].keys(): {}\n"
-                "descriptor['data_keys'].keys(): {}\n".format(
-                    data.keys(), timestamps.keys(), descriptor["data_keys"].keys()
+    descriptor: EventDescriptor,
+    event_counters: dict,
```
coretl, line 2301: `Dict[str, int]`? Ditto for other places this is mentioned

event_model/__init__.py
```diff
@@ -1797,353 +1872,647 @@ def __iter__(self):
...
+    data: dict,
+    timestamps: dict,
+    seq_num: List,
```
coretl, line 2304: `List[int]`?

coretl (approved):
Assuming there are no code changes, and this is just moving the code into dataclasses, then this looks fine to me

PR #264, opened 2023-05-16
Introduced the skeleton method of versioning

### LINKED ISSUE #263: Use the python skeleton's version determination method
The python3-pip-skeleton uses a much [simpler method](https://github.com/DiamondLightSource/python3-pip-skeleton/blob/main/src/python3_pip_skeleton/__init__.py) of versioning than `event-model` does  currently with `_version.py`. We should replace this method with the skeleton's.

evalott100:
Closes #263 

Changes to using the importlib metadata method of versioning.


event_model/__init__.py
```diff
@@ -1759,9 +1763,13 @@ def _is_array(checker, instance):
         for name, schema in schemas.items()
     }
 
+if sys.version_info < (3, 8):
+    from importlib_metadata import version  # noqa
+else:
+    from importlib.metadata import version  # noqa
```
coretl, line 1769: I'd put this in with the other importlib import at the top of the file

prjemian: Both branches call the same import.  Why is there an `if..else` clause?

prjemian: Removed my comment.  The replacement of `_` by `.` is subtle.

event_model/__init__.py
```diff
@@ -22,14 +22,21 @@
...
-from ._version import get_versions
-
 if sys.version_info < (3, 8):
-    from importlib_metadata import metadata
+    from importlib_metadata import metadata, version  # noqa
 else:
-    from importlib.metadata import metadata
+    from importlib.metadata import metadata, version  # noqa
+
+__version__ = version("event-model")
```
coretl, line 30: I would put the del here too:

```suggestion
__version__ = importlib_version("event-model")
del importlib_version
```

coretl (approved):
Do you still need the `#noqa`? Apart from that this looks fine.

evalott100:
Seems not, I'll get rid of it and merge

PR #269, opened 2023-05-24
Added unit test to test dots aren't accepted in keys

### LINKED ISSUE #268: Add a unit test for schema rejecting dots in keys
In bluesky, the test `test_documents.py::test_dots_not_allowed_in_keys` checks that the documents don't accept dots in keys. This test should be copied in event-model itself.

evalott100:
Closes #268 

Just copies the unit test from bluesky itself.

### DIFF
--- a/event_model/tests/test_auth.py
+++ b/event_model/tests/test_auth.py
@@ -1,7 +1,15 @@
+from uuid import uuid4
+
+import jsonschema
 import pytest
 from jsonschema.exceptions import ValidationError
 
 import event_model
+from event_model import DocumentNames, schema_validators
+
+
+def new_uid():
+    return str(uuid4())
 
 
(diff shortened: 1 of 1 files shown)


coretl (approved):
Looks fine to me. I vaguely recall @danielballan recounting why there might be random top level keys in the document, and why they shouldn't have dots in them, but I can't remember the details. This test seems to check that they are banned anyway.

danielballan:
Haha, I do recount many things on those calls. To put it in writing for posterity:

* We intend to allow extension of Event Descriptor ("random keys") without changes to the schema.
* But dots-in-keys is disallowed by Mongo. In other systems it may be allowed, but confusing. So we ban it. This is something that could be revisited/relaxed in the future if needed.

PR #270, opened 2023-05-25
Rethinking external assets

### LINKED ISSUE #236: Rethinking of external assets
Notes from a chat with @coretl @tacaswell @callumforrester @DiamondJoseph @tizayi 

The motivation here is to improve the situation for using the same detector in fly scans and step scans, with efficient access to data. Lots more to discuss before we implement these disruptive ideas, so nobody panic. :- )

<hr />

* Resource and Datum documents will be deprecated (!).
* As now, the `describe()` method continues to indicate which keys are backed by external data. (Perhaps we use a new word for this new scheme.)
* The `read()` method will simply _omit_ keys that are backed by external data. The RunBundler can verify that non-external keys are present and external keys (as declared by descriptor) are not present.
* If there are any external keys, RunBundler calls `collect_asset_docs()`. This should include a `Partition` document with some parameters that will be used to associcate a slice of underlying data with a (length-1) slice of Events.
* It should include a `Resource2` (gotta name this...) that should have a mimetype and a dict of arbitrary `parameters`. The only restrtiction on paramters is that they are jSON-serializable.
* The `Partition` will also have `index_start`, `index_stop` corresponding to a slice in the underlying storage.

```py
handler_class = registry[mimetype]  # i.e. `image/tiff` or `application/x-nexus-something`
handler = handler_class(**parameters)  # e.g. filename and whatever else goes here
# contrast to with datum: handler(**datum_kwargs)
handler[index_start:index_stop]
```

evalott100:
Closes #236


coretl (commented):
I think I'll review this once the other 2 are merged, too difficult to review at the moment

evalott100:
Changing to #272

PR #272, opened 2023-06-15
Rethinking external assets 2

### LINKED ISSUE #236: Rethinking of external assets

evalott100:
Closes #236 

Putting on a new branch for cleanliness

### DIFF
--- a/docs/user/explanations/data-model.rst
+++ b/docs/user/explanations/data-model.rst
@@ -477,31 +477,17 @@ See :doc:`external` for details on the role Stream Resource documents play in
 referencing external assets that are natively ragged, such as single-photon detectors,
 or assets where there are many relatively small data sets (e.g. scanned fluorescence data).
 
-Minimal nontrivial valid example:
-
-.. code-block:: python
-
-   # 'Stream Resource' document
-   {'path_semantics': 'posix',
-    'resource_kwargs': {},
-    'resource_path': '/local/path/subdirectory/data_file',
-    'root': '/local/path/',
-    'run_start': '10bf6945-4afd-43ca-af36-6ad8f3540bcd',
-    'spec': 'SOME_SPEC',
-    'stream_names': ['point_det'],
-    'uid': '272132cf-564f-428f-bf6b-149ee4287024'}
-
 Typical example:
 
 .. code-block:: python
 
-   # resource
-   {'spec': 'AD_HDF5',
+   # 'Stream Resource' document
+   {'data_key': 'detector_1',
+    'spec': 'AD_HDF5',
     'root': '/GPFS/DATA/Andor/',
     'resource_path': '2020/01/03/8ff08ff9-a2bf-48c3-8ff3-dcac0f309d7d.h5',
     'resource_kwargs': {'frame_per_point': 1},
     'path_semantics': 'posix',
-    'stream_names': ['point_det'],
     'uid': '3b300e6f-b431-4750-a635-5630d15c81a8',
     'run_start': '10bf6945-4afd-43ca-af36-6ad8f3540bcd'}
 
@@ -518,30 +504,17 @@ See :doc:`external` for details on the role Stream Datum documents play in refer
 external assets that are natively ragged, such as single-photon detectors,
 or assets where there are many relatively small data sets (e.g. scanned fluorescence data).
 
-Minimal nontrivial valid example:
-
-.. code-block:: python
-
-   # 'datum' document
-   {'resource': '272132cf-564f-428f-bf6b-149ee4287024',  # foreign key
-    'datum_kwargs': {},  # format-specific parameters
-    'datum_id': '272132cf-564f-428f-bf6b-149ee4287024/1',
-    'event_count': 1
-    }
-
 Typical example:
 
 .. code-block:: python
 
-   # datum
-   {'resource': '3b300e6f-b431-4750-a635-5630d15c81a8',
-    'datum_kwargs': {'index': 3},
-    'datum_id': '3b300e6f-b431-4750-a635-5630d15c81a8/3',
-    'event_count': 5,
-    'event_offset': 14}
-
-It is an implementation detail that ``datum_id`` is often formatted as
-``{resource}/{counter}`` but this should not be considered part of the schema.
+   # 'Stream Datum' document
+   {'uid': '86340942-9865-47f9-9a8d-bdaaab1bfce2',
+    'descriptor': '8c70b8c2-df32-40e3-9f50-29cda8142fa0',
+    'stream_resource': '272132cf-564f-428f-bf6b-149ee4287024',  # foreign key
+    'indices': {'start': 0, 'stop': 1},
+    'seq_nums': {'start': 1, 'stop': 2},
+    }
 
 Formal schema:
 
--- a/docs/user/explanations/data-model.rst
+++ b/docs/user/explanations/data-model.rst
@@ -526,7 +526,6 @@ Minimal nontrivial valid example:
    {'resource': '272132cf-564f-428f-bf6b-149ee4287024',  # foreign key
     'datum_kwargs': {},  # format-specific parameters
     'datum_id': '272132cf-564f-428f-bf6b-149ee4287024/1',
-    'block_idx': 0,
     'event_count': 1
     }
 
@@ -538,7 +537,6 @@ Typical example:
    {'resource': '3b300e6f-b431-4750-a635-5630d15c81a8',
     'datum_kwargs': {'index': 3},
     'datum_id': '3b300e6f-b431-4750-a635-5630d15c81a8/3',
-    'block_idx': 0,
     'event_count': 5,
     'event_offset': 14}
 
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1957,12 +1957,11 @@ class ComposeResource:
             root=root,
             resource_kwargs=resource_kwargs,
             resource_path=resource_path,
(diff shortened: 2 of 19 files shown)


event_model/documents/stream_datum.py
```diff
@@ -55,3 +41,24 @@ class StreamDatum(TypedDict):
...
     ]
+    data_keys: Annotated[
+        List[str],
+        Field(
+            description="A list to show which data_keys of the "
+            "Descriptor are being streamed"
+        ),
+    ]
+    seq_nums: Annotated[
+        Dict[str, int],
```
coretl, line 52: Could we make this another TypedDict rather than a generic Dict please?
```suggestion
        Range,
```
Where `Range` is:
```python
class Range(TypedDict):
    """The parameters required to describe a sequence of incrementing integers"""
    start: Annotated[
        int,
        Field(description="First number in the range"),
    ]
    stop: Annotated[
        int,
        Field(description="Last number in the range is less than this number"),
    ]
```

event_model/__init__.py
```diff
@@ -2035,18 +2032,21 @@ def __call__(
 def compose_stream_datum(
     *,
     stream_resource: StreamResource,
-    stream_name: str,
     counter: Iterator,
-    datum_kwargs: Dict[str, Any],
+    data_keys: List[str],
+    seq_nums: Dict[str, int],
+    indices: Dict[str, int],
```
coretl, line 2038: These should then both be `Range`

event_model/documents/stream_datum.py
```diff
@@ -18,13 +18,6 @@ class StreamDatum(TypedDict):
...
         ),
     ]
-    datum_kwargs: Annotated[
-        Dict[str, Any],
-        Field(
-            description="Arguments to pass to the Handler to retrieve one "
-            "quanta of data",
-        ),
-    ]
     event_count: Annotated[
```
coretl, line 21: `event_count` and `event_offset` can be deleted, they are now represented by `seq_nums`

event_model/documents/stream_datum.py
```diff
@@ -1,9 +1,23 @@
-from typing import Any, Dict
+from typing import List
 
 from typing_extensions import Annotated, TypedDict
 
 from .generate.type_wrapper import Field, add_extra_schema
 
+
+class Range(TypedDict):
```
coretl, line 8: Suggest `StreamRange`

event_model/__init__.py
```diff
@@ -2035,28 +2028,27 @@ def __call__(
...
-    event_offset: int = 0,
+    data_keys: List[str],
+    seq_nums: Range,
+    indices: Range,
     validate: bool = True,
 ) -> StreamDatum:
     """
     Here for backwards compatibility, the Compose class is prefered.
     """
-    return ComposeStreamDatum(stream_resource, stream_name, counter)(
```
coretl, line 2048: Add a warning that this will be removed next minor version

event_model/documents/stream_datum.py
```diff
@@ -18,40 +32,37 @@ class StreamDatum(TypedDict):
             "be contiguous for a given stream.",
```
coretl, line 32: @danielballan @tacaswell now we have `indices` I'm inclined to drop `block_idx` from `StreamDatum`, is that reasonable?

event_model/documents/stream_datum.py
```diff
@@ -18,40 +32,37 @@ class StreamDatum(TypedDict):
...
-    uid: Annotated[
-        str,
+    indices: Annotated[
+        StreamRange,
         Field(
-            description="Globally unique identifier for this Datum. A suggested "
-            "formatting being '<stream_resource>/<stream_name>/<block_id>",
+            description="A slice object passed to the StreamResource "
+            "handler so it can hand back data and timestamps"
         ),
```
coretl, line 67: @evalott100 we also need a pointer back to the Descriptor here (copy what we have in `Event`)

event_model/__init__.py
```diff
@@ -2000,31 +2000,24 @@ def compose_resource(
...
 class ComposeStreamDatum:
     stream_resource: StreamResource
-    stream_name: str
     counter: Iterator
 
     def __call__(
         self,
-        datum_kwargs: Dict[str, Any],
-        event_count: int = 1,
-        event_offset: int = 0,
```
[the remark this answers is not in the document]
coretl: Let's put `descriptor: Optional[EventDescriptor]` it in here, but default it to `None` so it can be created in the Ophyd Device without it, and the descriptor uid added later by bluesky

coretl (changes requested):
Just need that descriptor ref putting back in

event_model/__init__.py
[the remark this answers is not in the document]
evalott100: You think it should be in the `__call__` kwargs?  You mean adding an optional `descriptor_uid` string field to the `StreamDatum`?

event_model/documents/stream_datum.py
coretl: Yes it is, as we are going to want to squash adjacent StreamDatums

event_model/documents/stream_datum.py
```diff
@@ -1,57 +1,73 @@
...
 
 from typing_extensions import Annotated, TypedDict
 
 from .generate.type_wrapper import Field, add_extra_schema
 
+
+class StreamRange(TypedDict):
+    """The parameters required to describe a sequence of incrementing integers"""
+
+    start: Annotated[
```
coretl, line 11: I think I've found a use for `step`, we have an excalibur detector that writes round-robin to 4 HDF files, so would be good to say that indices `{40:80}` in fileA map to seq_nums `{161:321:4}`

event_model/__init__.py
```diff
@@ -1997,35 +1996,38 @@ def compose_resource(
     )
 
 
+# A dict of Tuple[str, StreamRange] where the string is the StreamDatum uuid
+
+_stream_datum_seq_nums: Dict[str, StreamRange] = {}
+
+
```
coretl, line 2003: I think this is unused?

event_model/__init__.py
```diff
@@ -2213,6 +2191,20 @@ def compose_stop(
     )(exit_status=exit_status, reason=reason, uid=uid, time=time, validate=validate)
 
 
+def dict_of_lists_has_equal_size_lists(dictionary: Dict[str, List]) -> bool:
```
coretl, line 2194: @danielballan do we support empty event pages? If so then we need to support an empty dict here.

coretl (commented):
One question occurred to me... Apart from that I really hope we're done! Thanks for sticking with this...

event_model/__init__.py
danielballan: Yes

danielballan:
I'm on board with retroactively applying `black`, but it would be best to do that in its own PR so that this PR is more reviewable as a unit.

evalott100:
> I'm on board with retroactively applying `black`, but it would be best to do that in its own PR so that this PR is more reviewable as a unit.

I was a little confused because the [CI was checking against black](https://github.com/bluesky/event-model/actions/runs/5548507016/job/15029844176#step:4:24). I think this relates to a local problem of line length being max 79 characters now (I think maybe flake8 changed the default line length, or the `max-line-length` option).

I'll revert the commit and see what's happening.

danielballan:
Oh, odd. I know we've retroactively applied `black` to some-but-not-all legacy bluesky repos. I can't remember for sure whether we did it here, but the single quotes in the codebase suggest we haven't.

evalott100:
@danielballan 

Found it!

https://github.com/bluesky/event-model/commit/01898f228dc365ceb766785120c707a9897eee36

I'm unsure why these things are changing now though... Maybe a new black release, I think I see some highlights in the recent black releases which match up (I used 23.9.1 locally) :thinking: 

I'll run black on a different branch off of main and make a new PR.

danielballan:
Ah, OK, that makes sense. I think we are good to merge then.

coretl (commented):
Merge when this is fixed

docs/user/explanations/data-model.rst
coretl, line 1: Please can you fix this example

coretl (approved):

PR #281, opened 2023-07-05
Fixed imports to avoid pydantic if the version is >=2

evalott100:
Will now use dummy imports if the pydantic version is >= 2.0.0, not only if pydantic doesn't exist.


event_model/documents/generate/type_wrapper.py
```diff
@@ -3,21 +3,39 @@
...
-try:
-    import pydantic
+pydantic_path = pkgutil.find_loader("pydantic")
+pydantic_version = None
+if pydantic_path:
+    import pydantic  # type: ignore
 
+    pydantic_version = pydantic.__version__
+
+if pydantic_version and pydantic_version < "2.0":
```
DominicOram, line 16: I think this is probably fine for the short term but could bite us in the future e.g.
```python
>>> "10.0">"2.0"
False
```

DominicOram:
From the docs it looks like another option might be in v2 you can do `from pydantic import v1 as pydantic`, which might mean you can get it to work for both?

event_model/documents/generate/type_wrapper.py
evalott100: Hopefully we'll convert to pydantic 2 relatively soon, but I just fixed this just in case.

evalott100:
> From the docs it looks like another option might be in v2 you can do `from pydantic import v1 as pydantic`, which might mean you can get it to work for both?

I've used try except with v1 now, so the generation code will work with either.

coretl:
> > From the docs it looks like another option might be in v2 you can do `from pydantic import v1 as pydantic`, which might mean you can get it to work for both?
> 
> I've used try except with v1 now, so the generation code will work with either.

But the upper cap is still in to stop pytdantic < 2?

event_model/documents/generate/typeddict_to_schema.py
```diff
@@ -33,13 +31,28 @@
...
 
+if not pydantic_version:
+    raise EnvironmentError(
+        "schema generation requires pydantic < 2.0 to run, pydantic isn't installed"
+    )
+elif int(pydantic_version.split(".")[0]) >= 2:
+    raise EnvironmentError(
+        "schema generation requires pydantic < 2.0 to run, current version "
+        f"{pydantic_version}"
+    )
```
DominicOram, line 55: Do we need this now if we're doing the v1 thing?

evalott100:
> > > From the docs it looks like another option might be in v2 you can do `from pydantic import v1 as pydantic`, which might mean you can get it to work for both?
> > 
> > 
> > I've used try except with v1 now, so the generation code will work with either.
> 
> But the upper cap is still in to stop pydantic < 2 being installed?

I think it's probably still good to pin pydantic<2 as a dev requirement since the generation code does require that version, the difference is that if event-model is installed as a requirement in an environment with pydantic 2 then the wrapper will use pydantic 1 instead.

event_model/documents/generate/typeddict_to_schema.py
evalott100: Actually, I'll leave in the not pydantic_version. I think it's beneficial to have in case someone tries to touch the generation code from an environment without pip, or from event-model without dev requirements.

coretl (approved):

PR #286, opened 2023-08-23
Add ability for ComposeDescriptor to accept arbitrary kwargs to create descriptor

rosesyrett:
## Description
Currently the DAQ-Core team at diamond are in need of being able to pass metadata to descriptor documents, primarily for nexus file writing. 

## Motivation and Context
In our case, we want descriptor documents from streams to be easily identifiable and match up with the Application definitions for the experiments they represent (e.g. for [tomography,](https://manual.nexusformat.org/classes/applications/NXtomo.html) detectors require an image key `DataKey` field).

## How Has This Been Tested?
I've written a test confirming that extra kwargs are passed to the resulting descriptor document.

### DIFF
--- a/event_model/tests/test_schema_generation.py
+++ b/event_model/tests/test_schema_generation.py
@@ -6,6 +6,7 @@ from time import time
 import pytest
 
 from event_model import ComposeDescriptor
+from event_model.documents import RunStart
 from event_model.documents.event_descriptor import DataKey
 
 if sys.version_info[:2] >= (3, 9):
@@ -21,7 +22,6 @@ if sys.version_info[:2] >= (3, 9):
         EventDescriptor,
         EventPage,
         Resource,
-        RunStart,
         RunStop,
         StreamDatum,
         StreamResource,


evalott100:
I'd be for adding this functionality to the `ComposeDescriptor` dataclass, where `EventDescriptor` is initialised:

https://github.com/bluesky/event-model/blob/ef8ee4b6040e647883d8adbe6c5fa9da33ac984a/event_model/__init__.py#L2417-L2426

I'd suggest adding a `metadata` kwarg  to the `ComposeDescriptor::__call__` then if metadata is passed in after the `EventDescriptor` typeddict is initialised we can `.update` with the metadata if it exists, all before schema validation

evalott100:
> At the moment I'm seeing mypy errors:
> 
> ```
> event_model/__init__.py:2427: error: Unsupported type "dict[str, Any]" for ** expansion in TypedDict  [typeddict-item]
> Found 1 error in 1 file (checked 4 source files)
> ```
> 
> This is because I'm trying to unpack `kwargs` into the EventDescriptor, which is a TypedDict. The JSON schema suggests we can have any arbitrary fields, but the python TypedDict is obviously restrictive in what it can handle. Is there any nice way to deal with this?

@RAYemelyanova After the `EventDescriptor`TypedDict is initialised, but before the schema validation, I'd suggest doing a 
`doc.update(kwargs)`. This is how it's currently handled in `compose_run`.

event_model/__init__.py
```diff
@@ -2402,6 +2402,7 @@ def __call__(
         time=None,
         uid=None,
         validate=True,
+        **kwargs,
```
evalott100, line 2405: I think we should handle this the same way as RunStart (without the unpacking)

https://github.com/bluesky/event-model/blob/ef8ee4b6040e647883d8adbe6c5fa9da33ac984a/event_model/__init__.py#L2512

I'm happy to change it to unpacking if others prefer, but I think they should be consistent

event_model/__init__.py
```diff
@@ -2423,6 +2424,7 @@ def __call__(
             time=time,
             uid=uid,
             hints=hints,
+            **kwargs,
```
evalott100, line 2427: If we do the `doc.update(kwargs)` I mentioned above then maybe we should write a check that the metadata doesn't have any keys matching the required or optional ones in the TypedDict.

evalott100: Additionally could we change `compose_run` to create a `RunStart` directly and add the metadata afterwards? I probably should have done this in my initial conversion to TypedDicts

evalott100 (commented):
Looks good, I don't think there'll be a mypy compliment way to add any possible key from `kwargs` into the `__init__` of the typed dict, I'd suggest updating the dict after it's been initialised, but I'm open to whatever

evalott100:
I deleted some previous comments on this, they also result in mypy errors down the line... Should maybe discuss solutions at the meeting

evalott100:
This passes mypy on event-model:
``` python
# =================== __init__.py
class ComposeDescriptor:
    start: RunStart
    streams: dict
    event_counters: Dict[str, int]

    def __call__(
        self,
        name,
        data_keys,
        hints=None,
        configuration=None,
        object_keys=None,
        time=None,
        uid=None,
        validate=True,
        metadata=None,
    ) -> ComposeDescriptorBundle:
        if time is None:
            time = ttime.time()
        if uid is None:
            uid = str(uuid.uuid4())
        if hints is None:
            hints = {}
        if configuration is None:
            configuration = {}
        if object_keys is None:
            object_keys = {}

        doc = EventDescriptor(
            configuration=configuration,
            data_keys=data_keys,
            name=name,
            object_keys=object_keys,
            run_start=self.start["uid"],
            time=time,
            uid=uid,
            hints=hints,
        )
        if metadata:
            doc.update(metadata)

        if validate:
            if name in self.streams and self.streams[name] != set(data_keys):
                raise EventModelValidationError(
                    "A descriptor with the name {} has already been composed with "
                    "data_keys {}. The requested data_keys were {}. All "
                    "descriptors in a given stream must have the same "
                    "data_keys.".format(name, self.streams[name], set(data_keys))
                )
            schema_validators[DocumentNames.descriptor].validate(doc)

        if name not in self.streams:
            self.streams[name] = set(data_keys)
            self.event_counters[name] = 1

        return ComposeDescriptorBundle(
            descriptor_doc=doc,
            compose_event=ComposeEvent(
                descriptor=doc, event_counters=self.event_counters
            ),
            compose_event_page=ComposeEventPage(
                descriptor=doc, event_counters=self.event_counters
            ),
        )

# =================== test.py
run_doc, compose_descriptor, compose_resource, compose_stop = compose_run()

descriptor_doc, compose_event, compose_event_page = compose_descriptor(
    "name", {}, metadata={"a": "b"}, validate=True
)


descriptor_doc["data_keys"]
descriptor_doc["run_start"]
descriptor_doc["a"]


def foo(descriptor_doc: EventDescriptor):
    ...

foo(descriptor_doc)

PR #289, opened 2023-10-10
Mnt py312

tacaswell:
## Description

Drop support for old jsonschema (to drop need to parse version strings)

## Motivation and Context

Looking at https://github.com/conda-forge/event-model-feedstock/pull/46 the failures are because our tests use `distutils` which were removed in py312.  The first commit changes to use `packgaing.version.parse` instead.  However, we only need to check version strings to support jsonschema < 3 which at this point is almost 5 years old so I think we can safely require a newer version of `jsonschema`.

If changing the dep pinning in too aggressive I can drop the second commit.

attn @ZLLentz

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -5,7 +5,7 @@ on:
   pull_request:
 env:
   # The target python version, which must match the Dockerfile version
-  CONTAINER_PYTHON: "3.11"
+  CONTAINER_PYTHON: "3.12"
 
 jobs:
   lint:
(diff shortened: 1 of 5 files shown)


danielballan (approved):

danielballan (changes requested):
Upon closer inspection, this PR needs to remove the CI jobs targeting jsonschema 2.x

tacaswell:
I 👍🏻 these changes, but can not in the UI because it is my own PR.

danielballan (approved):

danielballan:
Thanks @untzag !

PR #294, opened 2024-02-01
Update pydantic to v2

evalott100:
## Description
Changed event model to use pydantic v2. This is now pinned. 

## Motivation and Context
Since the release of pydantic v2, we've been doing generation code with pydantic 1. Pydantic 2 allows us to do this with less code.

## How Has This Been Tested?
Tested against bluesky itself. So long as the schema match up the runtime code should work the same.


pyproject.toml
```diff
@@ -51,7 +51,7 @@ dev = [
     "numpydoc",
 
     # For schema generation.
-    "pydantic<2.0",
+    "pydantic~=2.6.0",
```
danielballan, line 54: In library code, as opposed to application code, we generally avoid pinning things _down_, as this can create more problems than it solves. (Gallons of proverbial blog post ink have been spilled on this subject: [here's one](https://iscinumpy.dev/post/bound-version-constraints/) that links out to others if anyone wants to fall down the rabbithole.)

evalott100: Good post! I'll pin >= 2.6

event_model/documents/generate/type_wrapper.py
```diff
@@ -2,57 +2,92 @@
...
+def Field(*args, **kwargs):  # type: ignore
+    ...
 
-    def Field(*args, **kwargs):  # type: ignore
-        ...
 
-    class FieldInfo:  # type: ignore
-        ...
+class BaseModel:  # type: ignore
+    def __init__(self, *args, **kwargs): ...
```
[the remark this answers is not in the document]
evalott100: Accidentally left in from when I was calling creating an instance of `BaseModel` instead of `TypeAdapter`

danielballan (approved):

PR #297, opened 2024-03-06
Fix minor bug in event_model's rechunk_event_pages

cjtitus:
`rechunk_event_pages` has a bug if you try to re-chunk an EventPage that has an empty "filled" key, because the dictionary comprehension to create the new "filled" dictionaries uses keys from "data". This means that you can't rechunk the EventPages that are emitted by the RunEngine.

## Description

I have just changed one word -- instead of 
`"filled": {key: page["filled"][key][start:stop] for key in page["data"].keys()},`
I have written
`"filled": {key: page["filled"][key][start:stop] for key in page["filled"].keys()},`
which correctly leaves filled empty, if it was empty to begin with.

## Motivation and Context

It is necessary to fix this bug in order to re-chunk EventPages that are currently too large to go into Kafka after long fly-scans.

## How Has This Been Tested?

I have tested this by manually calling rechunk_event_pages on some generated EventPages, and also by running the built-in tests.

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -2772,7 +2772,7 @@ def rechunk_event_pages(event_pages: Iterable, chunk_size: int) -> Generator:
                     for key in page["timestamps"].keys()
                 },
                 "filled": {
-                    key: page["filled"][key][start:stop] for key in page["data"].keys()
+                    key: page["filled"][key][start:stop] for key in page["filled"].keys()
                 },
             }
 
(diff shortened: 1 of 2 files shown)


tacaswell (approved):

cjtitus:
Ahh, I spoke too soon! It turns out that rechunk_event_pages still fails if you _increase_ the chunk size. So, if you input a list of EventPages and want to increase the chunk size to decrease the number of pages, the constructor for EventPages also needs to have a non-empty "filled" key.

I need to fix this as well, so don't merge yet.

cjtitus:
Ok, now I can chunk up or down. I wonder if some tests should be implemented for this...

mrakitin (changes requested):
Thank you for the fix, @cjtitus.

Please fix the formatting which became broken due to the fix.

If you could add a simple test to demonstrate the bug and its fix, it would be great!

danielballan (approved):

PR #312, opened 2024-08-20
ENH: Added NX_class as a PerObjectHint to the event descriptor.

burkeds:
This is an optional hint that can be used to carry a NeXus class definition one may wish to associate with a device.

## Motivation and Context
It would be good to have a standard way to assign a NeXus class definition to a high level device. Having a specific field with a standard name will better allow the community to converge on a standard way to assign and use this definition for high level processes such as writing callbacks.

### DIFF
--- a/event_model/documents/event_descriptor.py
+++ b/event_model/documents/event_descriptor.py
@@ -69,6 +69,15 @@ class PerObjectHint(TypedDict):
             Field(description="The 'interesting' data keys for this device."),
         ]
     ]
+    NX_class: NotRequired[
+        Annotated[
+            str,
+            Field(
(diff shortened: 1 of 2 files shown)


danielballan (approved):
This looks good to me, but @evalott100 should sign off before we merge, to confirm that there's nothing special we have to do with schema generation here.

evalott100:
> there's nothing special we have to do with schema generation here

CI does our generation check! LGTM :pray:

evalott100 (approved):

danielballan:
Nice to have a @burkeds commit in event-model. Thanks Devin!

PR #313, opened 2024-09-17
removed `Optional` on the specified datakey fields

### LINKED ISSUE #309: Add NotRequired things to the DataKey schema
As per https://github.com/bluesky/ophyd-async/pull/137#discussion_r1624442103

Also removing the `Optional` from "units" and "precision"

https://github.com/bluesky/event-model/blob/fc6e30eae896073eda4013f7c4af3362fead61c1/event_model/documents/event_descriptor.py#L59

evalott100:
Fixes half of #309

### DIFF
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -34,6 +34,7 @@ import jsonschema
 import numpy
 from typing_extensions import Literal
 
+from .documents import Dtype
 from .documents.datum import Datum
 from .documents.datum_page import DatumPage
 from .documents.event import Event
@@ -61,6 +62,7 @@ del importlib_version
 
 
 __all__ = [
+    "Dtype",
     "DocumentNames",
     "schemas",
     "schema_validators",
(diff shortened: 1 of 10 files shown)


coretl (approved):
https://github.com/bluesky/event-model/issues/309 isn't a very well written ticket... This PR does half of it.

Please could you make another PR that adds `limits` and `choices` as per the link in the ticket and make them `NotRequired` fields.

DiamondJoseph (commented):
Aren't units and precision optional though?
For a str enum precision doesn't make sense- are we just going to say 0 or -1? Empty units for dimensionless things yes

coretl:
> Aren't units and precision optional though? For a str enum precision doesn't make sense- are we just going to say 0 or -1? Empty units for dimensionless things yes

We remove the ability to make them `null`. This means they don't need to be in the descriptor (e.g. `precision` for an enum), but if they **are** present they must be of the right type.

DiamondJoseph (approved):
That makes sense: precision is still NotRequired but if present is str not str | None

evalott100:
> Please could you make another PR that adds `limits` and `choices` as per the link in the ticket and make them `NotRequired` fields.

May as well do it here

evalott100:
@coretl I think I should maybe make each field `Optional` in `LimitsRange` (as opposed to `NotRequired`). EPICS does allow for low/high limits without high/low being set.

Issue #308, opened 2024-05-30
Add specific data type


danielballan:
EventDescriptors specific that data type of the data in Events and StreamResources in the key `dtype` using [jsonschema datatypes](https://cswr.github.io/JsonSchema/spec/basic_types/):

https://github.com/bluesky/event-model/blob/145118722a2e2f15abf32010e3e6b71506398140/event_model/schemas/event_descriptor.json#L47-L53

This is the tragedy of, "You must define the core of your software at the beginning, when you understand the problem the least." At the time (2015) we were focused on MongoDB (bson) and Python applications, where data types can be coarsely defined. We now view this as a mistake. The `array` option is particular does not make sense: we have `shape` for that. We should have given specific types.

How should we add them now?

## Decision: New key or expand `dtype` enum?

If we expand the `dtype` enum to optionally specify a specific data type instead of the jsonschema types, this could break downstream consumers (some in code that we do not know about) that have been able to expect the jsonschema types for the last ~8 years. It seems safest to add a new key sitting beside `dtype`. In "Bluesky 2.0" this could be cleaned up / consolidated and documented as a backward-incompatible change.

## Decision: How to spell the data type?

Three ideas have been proposed:

1. Use the [NumPy array protocol type string (typestr) format](https://numpy.org/doc/stable/reference/arrays.interface.html#the-array-interface-protocol), e.g. `"<f8"`, `">i4"`, `"|b1"`. There is precedent for using this as a way of encoding numpy data types in JSON: the [Zarr v2 spec](https://zarr-specs.readthedocs.io/en/latest/v2/v2.0.html#data-type-encoding) does so.
2. Use the newer [Zarr v3 specification](https://github.com/zarr-developers/zarr-specs/blob/a6f198040571557484c29dac34d3f37ceeae7f9f/docs/v3/core/v3.0.rst#core-data-types), which opts for a more constrained set of supported types with more human-readable names, e.g. `float64`, `int8`, `bool`. Types are little-endian. Big-endianness is a handled as a property of the encoding (a codec).
3. Use Arrow, which supports a super-set of these types. However, Arrow has no officially-supported JSON encoding. Its schema is binary; it would be have to be base64-encoded or similar---not human-readable. For that reason, I think it is easy to reject this option.

[...]

PR #315, opened 2024-09-20
Add `dtype_numpy` with medium-effort validation

danielballan:
Add `dtype_numpy` to Event Descriptor.

Supersedes #215.

PR #316, opened 2024-09-30
Dtype numpy

evalott100:
Replacing https://github.com/bluesky/event-model/pull/315

### DIFF
--- a/.github/actions/install_requirements/action.yml
+++ b/.github/actions/install_requirements/action.yml
@@ -12,7 +12,7 @@ inputs:
     default: "3.x"
   jsonschema_version:
     description: version of the jsonschema pip package to install
-    default: 3
+    default: 4
 
 runs:
   using: composite
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -40,7 +40,7 @@ jobs:
           - os: "ubuntu-latest"
             python: "3.8"
             install: ".[dev]"
-            jsonschema: 3
+            jsonschema: 4
 
     runs-on: ${{ matrix.os }}
     env:
--- a/event_model/__init__.py
+++ b/event_model/__init__.py
@@ -1801,18 +1801,20 @@ for name, filename in SCHEMA_NAMES.items():
 
 def _is_array(checker, instance):
     return (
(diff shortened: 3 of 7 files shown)


evalott100:
@coretl @danielballan 

Despite the fact pydantic will fail correctly for the list field here:

https://github.com/bluesky/event-model/blob/94c7801d1d42a4ccff9da82d24721ec35d922dd3/event_model/tests/test_auth.py#L78-L88

The schema will not, it's because for some reason pydantic is outputting `prefixItems` instead of just `items` here:

https://github.com/bluesky/event-model/blob/94c7801d1d42a4ccff9da82d24721ec35d922dd3/event_model/schemas/event_descriptor.json#L76

evalott100:
Inserting `"items": false` makes everything work, I'm not too sure why pydantic isn't doing this already.

``` json
                {
                            "items": {
                                "maxItems": 2,
                                "minItems": 2,
                                "prefixItems": [
                                    {
                                        "type": "string"
                                    },
                                    {
                                        "description": "A numpy dtype e.g `<U9`, `<f16`",
                                        "pattern": "[|<>][tbiufcmMOSUV][0-9]+",
                                        "type": "string"
                                    }
                                ],
                                "items": false,
                                "type": "array"
                            },
                            "type": "array"
```

evalott100:
For now we'll get around this by adding

```
_ConstrainedDtypeNpStructure = Annotated[
    Tuple[str, _ConstrainedDtype], Field(json_schema_extra={"items": False})
]
```

But I'll make an issue on pydantic.

evalott100:
Okay I figured it out, we simply can't pass a tuple into the jsonschema... While it will do positional args differently, it requires the `array` to be a list... But then paradoxically to generate the correct schema which does have different constraints on `prefixItems` pydantic will need the type hint to be `Tuple`.

evalott100:
@tacaswell implemented this to allow for tuples, but it isn't working in this case

https://github.com/python-jsonschema/jsonschema/issues/148#issuecomment-2058122215

coretl (approved):

coretl:
fixes #308

PR #318, opened 2024-10-02
v1.22.0 release notes

### LINKED ISSUE #317: Draft release notes and expose `event_descriptor` `Limits`/`LimitsRange`
We'll do a new minor release, since both the above should be quick to review we'll do them both in the same PR.

evalott100:
Closes #317


evalott100:
I think we'd quite like to make another minor release. Are we okay to do this?

image

Things which could be breaking:
* If anything downstream can't also upgrade jsonschema (we require this to validate tuples index-wise).
* If anything is passing `None` directly into `precision`, `choices`, or `units`.

evalott100:
image

coretl (approved):

PR #332, opened 2024-10-30
license README badge updated to BSD

evalott100:

PR #333, opened 2024-10-30
Update README.md

coretl:
- Fix license
- Fix title
- Fix logo


mrakitin (approved):
Looks like it's overriding #332, correct?
In that PR, `https://bluesky.github.io/event-model` was wrapped into angle braces in the `README.md` file (not sure if it's needed).

evalott100:
> not sure if it's needed

It's what my linter preferred for links unembedded in words, but it doesn't make a difference in this case.

evalott100 (approved):

PR #341, opened 2024-12-17
Convert from `BaseModel` to static jsonschema + `TypedDict`

### LINKED ISSUE #337: BaseModel documents and change to schema generation
There's been some demand for pydantic `BaseModel` versions of the documents.

I propose we change event-model document generation to allow for these, in a backwards compatible way.

## 1: Converting the current jsonschema to pydantic models

Most `TypedDict` document definitions only add `{"additionalProperties": False}` to the outputted schema, which is implicit in pydantic models, so most pydantic documents will be identical to the current ones swapping out `TypedDict` for `BaseModel`. There are other places where we add more complex logic to the schema.

### Run Stop

In run-stop we have the following extra schema:

``` python
RUN_STOP_EXTRA_SCHEMA = {
    "patternProperties": {"^([^./]+)$": {"$ref": "#/$defs/DataType"}},
    "additionalProperties": False,
}
```

Which we can represent in pydantic as:

``` python
class RunStop(BaseModel):
    data_type: DataType
    # ... other non `DataType` fields

    class Config:
        extra = 'allow'

    @root_validator(pre=True)
    def validate_additional_fields(cls, values):
        for key, value in values.items():
            if '.' not in key and key not in cls.__fields__:
                try:
                    datatype = DataType.parse_raw(value)
                    setattr(self, key, datatype)
                except ValidationError as err:
                    raise ValueError(f"Extra non-datatype {key} received.") from err
        return values
```

### Event Descriptor

In event-descriptor we have the following extra schema:

``` python
EVENT_DESCRIPTOR_EXTRA_SCHEMA = {
    "patternProperties": {"^([^./]+)$": {"$ref": "#/$defs/DataType"}},
    "$defs": {
        "DataType": {
            "title": "DataType",
            "patternProperties": {"^([^./]+)$": {"$ref": "#/$defs/DataType"}},
            "additionalProperties": False,
        },
    },
    "additionalProperties": False,
}
```

Which we can represent in pydantic the same way as above.

### Run Start

The run-start additional schema is substantially more complicated:

``` python
RUN_START_EXTRA_SCHEMA = {
    "$defs": {
        "DataType": {
            "patternProperties": {"^([^./]+)$": {"$ref": "#/$defs/DataType"}},
            "additionalProperties": False,
        },
        "Projection": {
            "allOf": [
                {
                    "if": {
                        "allOf": [
                            {"properties": {"location": {"enum": ["configuration"]}}},
                            {"properties": {"type": {"enum": ["linked"]}}},
                        ]
                    },
                    "then": {
                        "required": [
                            "type",
                            "location",
                            "config_index",
                            "config_device",
                            "field",
                            "stream",
                        ]
                    },
                },
                {
                    "if": {
                        "allOf": [
                            {"properties": {"location": {"enum": ["event"]}}},
                            {"properties": {"type": {"enum": ["linked"]}}},
                        ]
                    },
                    "then": {"required": ["type", "location", "field", "stream"]},
                },
                {
                    "if": {
                        "allOf": [
                            {"properties": {"location": {"enum": ["event"]}}},
                            {"properties": {"type": {"enum": ["calculated"]}}},
                        ]
                    },
                    "then": {"required": ["type", "field", "stream", "calculation"]},
                },
                {
                    "if": {"properties": {"type": {"enum": ["static"]}}},
                    "then": {"required": ["type", "value"]},
                },
            ],
        },
    },
    "properties": {
        "hints": {
            "additionalProperties": False,
            "patternProperties": {"^([^.]+)$": {"$ref": "#/$defs/DataType"}},
        },
    },
    "patternProperties": {"^([^./]+)$": {"$ref": "#/$defs/DataType"}},
    "additionalProperties": False,
}
```

1. The `DataType` `root_validator` can be added to the `Hints` and `RunStart` as above.

2. For `Projections` the sanest way to adjust what we have currently would be to create a new model for each projection type and then add them as a union in `RunStart`, this would have the effect of defining a couple of different `Projection` types in the outputted schema, though it wouldn't be breaking. Alternatively there's the following method:

``` python
class Projection(BaseModel):
    type: Literal['linked', 'calculated', 'static']
    location: Optional[Literal['configuration', 'event']] = None
    config_index: Optional[int] = None
    config_device: Optional[str] = None
    field: Optional[str] = None
    stream: Optional[str] = None
    calculation: Optional[str] = None
    value: Optional[str] = None

    @root_validator(pre=True)
    def check_required_fields(cls, values):
        type_ = values.get('type')
        location = values.get('location')

        if type_ == 'linked' and location == 'configuration':
            required_fields = ['type', 'location', 'config_index', 'config_device', 'field', 'stream']
        elif type_ == 'linked' and location == 'event':
            required_fields = ['type', 'location', 'field', 'stream']
        elif type_ == 'calculated' and location == 'event':
            required_fields = ['type', 'field', 'stream', 'calculation']
        elif type_ == 'static':
            required_fields = ['type', 'value']
        else:
            required_fields = []

        for field in required_fields:
            if values.get(field) is None:
                raise ValueError(f'{field} is required for type {type_} and location {location}')

        return values
```

## 2: Updating the schema generation

Currently, we generate the jsonschema from the `TypedDict` definitions with pydantic, and add the `EXTRA_SCHEMA` dictionaries.

Instead, we'll define the pydantic models, package the schema representation of the `root_validator`s within them and then generate both the jsonschema and the TypedDicts from the pydantic models (statically).

## 3: Optional fields

Pydantic doesn't allow for fields to be `NotRequired`, a field which is `NotRequired` in the `TypedDict` would have to be `None` in the pydantic model. For this reason we will forbid fields being `Optional` having a different meaning to `NotRequired`.

Fields which are `Optional` with default `None` in the `BaseModel` will be `NotRequired[Optional[...]]` in the `TypedDict`.

evalott100:
Closes #337


coretl (commented):
Neat! This is looking pretty good

src/event_model/schemas/bulk_events.json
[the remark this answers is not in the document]
evalott100: Happy for them to remain for now

src/event_model/documents/__init__.py
```diff
@@ -1,62 +1,13 @@
...
-    DatumPage,
-    Event,
-    EventDescriptor,
-    EventPage,
-    Resource,
-    RunStart,
-    RunStop,
-    StreamDatum,
-    StreamResource,
-)
```
coretl, line 44: I think this file probably has to go back to how it was, at least for the presence of `DocumentType` and `ALL_DOCUMENTS`

src/event_model/schemas/event_descriptor.json
```diff
@@ -10,22 +10,47 @@
...
-                    "type": "object"
+                    "anyOf": [
+                        {
+                            "type": "object"
+                        },
+                        {
+                            "type": "null"
+                        }
+                    ],
+                    "default": null
```
coretl, line 21: Hmm, this is actually saying that downstream clients now need to cope with the presence of this key, and its value being nul...

src/event_model/documents/event_descriptor.py
```diff
@@ -1,214 +1,168 @@
...
-    ]
-    data_keys: NotRequired[
-        Annotated[
-            Dict[str, DataKey],
-            Field(
-                description="This describes the data stored alongside it in this "
-                "configuration object."
-            ),
-        ]
-    ]
```
coretl, line 143: ...it used to be `NotRequired[Dict]`...

src/event_model/basemodels/event_descriptor.py
```diff
@@ -0,0 +1,238 @@
...
+            description="The NeXus class definition for this device.",
+            pattern=r"^NX[A-Za-z_]+$",
+        ),
+    ]
+
+
+class Configuration(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+    data: Annotated[
+        Union[Dict[str, Any], None],
```
coretl, line 156: ...and now it's `Union[Dict, None]`.

Is there any way to persuade pydantic to treat `None` as `NotRequired` in the JSON schema?

jacopoabramo: Sorry for intruding but I'm interested in the progress of this PR.

I checked around a bit and the behavior of `NotRequired` seems to be exclusive to `TypedDict`. I looked over and they seem to have asked a similar question. In this [issue](https://github.com/pydantic/pydantic/issues/1223) they made a [decorator](https://github.com/pydantic/pydantic/issues/1223#issuecomment-775363074) to use fields as "optional", but not in the sense of `Optional[Any]`  as in "it can either be `Any` or `None`" but rather "this field can be present or not", which I believe is more closely related to the behavior of `NotRequired` I imagine.

Not sure if it can be of help.

evalott100: @coretl 
> Is there any way to persuade pydantic to treat `None` as `NotRequired` in the JSON schema?

The jsonschema looks the same as before with one exception - on top of being absent from the `"required"` list, the `NotRequired` elements explicitly allow `"null"` as a datatype.

Since the default is also `"null"`, the converter does give us a `NotRequired` element in the `TypedDict`:

https://github.com/bluesky/event-model/blob/77689f6d1e881a453f4cd0386ac3e23d8bf17a9f/src/event_model/documents/event_descriptor.py#L120

This difference is necessary since pydantic doesn't natively support `NotRequired`. @jacopoabramo has linked to a method which supports this, but I don't think we should use:
* These examples convert the BaseModel to a dictionary before accessing members, because a `model.member` would give an error for a `NotRequired`. Whenever a user wants to actually do anything with their `BaseModel` they'll have to convert to and from a dictionary. Rather than just accessing members and checking if they're `None`.
* Pydantic supports excluding unset/`None` members during serialization anyway, and I'd prefer to not deviate too much from the vanilla api in the document structures themselves.

The downside to all of this is that in `TypedDict` `None` values will be allowed and will have the same meaning as the element being absent.

evalott100: Actually, we could get around all of this. The datamodel-codegen has options to generate a `BaseModel`, or a `TypedDict` from `jsonschema`, for a non-"null" not required jsonschema field it will do `Optional`, `default=None` for the former, and simply `NotRequired` for the latter.

I think this is nicer, will implement today.

evalott100:
@jacopoabramo

Some of the schema will allow additional fields if the given field has a key which doesn't contain dots. This means some schema, e.g `run_start` or `event_descriptor`, will be of variable length. I think this may make them incompatible with msgspec, which I assume optimizes based on fixed definitions for keys?

jacopoabramo:
> I think this may make them incompatible with msgspec, which I assume optimizes based on fixed definitions for keys?

Not necessarely but right off the bat I can't be certain. I wanted to try it out once this PR is completed, so to also review it myself to replicate the same steps.

pyproject.toml
```diff
@@ -50,11 +50,12 @@ dev = [
...
 
   # For schema generation.
-  "pydantic>=2.6",
+  "pydantic<3",
+  "datamodel-code-generator",
 ]
 
 [project.scripts]
-regenerate-schema = "event_model.documents.generate.__main__:main"
+regenerate-documents = "event_model.generate.__main__:generate"
```
coretl, line 58: I wonder if we should be exposing this as a script, as it's only used internally to the module. Could we ditch this section and call this via `python -m event_model.generate` instead?

coretl (commented):
Did I get this right, the source is now the JSON schemas rather than TypedDicts or BaseModels? Hmm, I'm not a big fan of writing JSON schemas by hand. Let's discuss offline...

PR #343, opened 2025-01-10
Convert from `BaseModel` to `jsonschema` + `TypedDict`

### LINKED ISSUE #337: BaseModel documents and change to schema generation

evalott100:
Closes #337


evalott100:
Will probably streamline the system of adding extra json schema to BaseModels too

coretl (commented):
I'm happy with this, and it didn't require any code changes so I think that means the checking for nulls is OK.

@danielballan please can you review the schema changes on projections?

Just to make sure, please can you point bluesky to this and run `mypy src/bluesky`, then point ophyd-async at this and run `pyright src/ophyd_async` and check it doesn't think we need to check for nulls anywhere?

coretl:
Also, why is the test failing?

evalott100:
I tested that it's working against bluesky and ophyd-async, both tests and type checking.

I had to write some custom code to include `Partial` documents within the jsonschema itself:

https://github.com/bluesky/event-model/blob/2195892117a62c81b28fb579022b92ff060ae1ab/src/event_model/generate/create_documents.py#L169

danielballan (approved):
Looks good.

Incidentally I like the new usage, too. Good use of `python -m ...`.

coretl (approved):

evalott100:
Was going to add tests for the `BaseModel` themselves. I think it's better to do it in a different PR though.

PR #346, opened 2025-01-15
Add py.typed file to package

thomashopkins32:
## Description

Add `py.typed` marker for the package.

## Motivation and Context

When using ophyd-async, the mypy extension in VS Code complains that there is no py.typed stub in event-model.


thomashopkins32:
@evalott100 Ah I'm guessing because bluesky has typing errors related to event-model objects?

evalott100:
> Ah I'm guessing because bluesky has typing errors related to event-model objects?

It's because we have some `TypedDict` e.g `datum_doc` which we know is `Datum`, but type checking sees as some union of documents because it comes from a structure typed as such.

https://github.com/bluesky/bluesky/blob/ca4f8aa136c91dccd428936d2d0ebe963b925f4e/src/bluesky/callbacks/tiled_writer.py#L197-L202

Here `_docs_cache` values are typed as `Datum | Resource | StreamResource` so you get key errors on `"datum_id"` for those possibilities. I would see if you can clean up the typing, though I think most of these need to be handled with casting, i.e

``` python
                    datum_doc = typing.cast(Datum, self._docs_cache.pop(datum_id))
```

since there's no way the type checker will know that an element with `uid` `datum_id` has to be a `Datum`.

thomashopkins32:
Okay, I'll try to take a stab at it and see if I can get it to work. Thanks for the info!

evalott100:
> Okay, I'll try to take a stab at it and see if I can get it to work. Thanks for the info!

The above example can be solved by making individual `_datum_docs_cache`, `_stream_resource_docs_cache` and `_resource_docs_cache` instead of `cast`ing, then add to the respective cache in these:

https://github.com/bluesky/bluesky/blob/ca4f8aa136c91dccd428936d2d0ebe963b925f4e/src/bluesky/callbacks/tiled_writer.py#L212-L219

This would be nicer. For some things you might have to cast instead though.

thomashopkins32:
@evalott100 I finished the changes to bluesky that are blocking this PR from being merged. Please see https://github.com/bluesky/bluesky/pull/1873

I tried my best not to simply cast where I could but indeed some places it was necessary.

evalott100 (approved):
[Bluesky changes to accomodate](https://github.com/bluesky/bluesky/pull/1873) this ready to be merged as well.

PR #347, opened 2025-03-06
Fix EPICS enum signals support

flowln:
In EPICS, enum (`mbbi` / `mbbo`) values [do not possess units](https://github.com/epics-base/epics-base/blob/13d6ca598cca495b2e559e808392ed22265c57e3/modules/ca/src/client/db_access.h#L459), so Ophyd [fills their values with `None`](https://github.com/bluesky/ophyd/blob/4a14425389f1e46b2c332e954c791bf0ae0cb175/ophyd/signal.py#L217) before composing the descriptor document.

As such, any such signals would fail validation prior to this patch.

This reverts a behavior change from PR https://github.com/bluesky/event-model/pull/343.


danielballan:
Thanks!

The change to the schema seems OK, but I have question about pinning down the dependency. From the commit message:

> Since we're using PythonVersion.PY_38 in the typeddict generator,
> we can't use the newer versions, that dropped support for Python
> version older than 3.9.

This kind of thing does not generally require any action on our part. Package metadata on each release of `datamodel-code-generator` should indicate which versions of Python it supports, and pip should get the most recent release that supports the appropriate Python version. So, in a Python 3.8 environment, we would get the latest release that supported 3.8, and in a Python 3.9+ environment, we would get potentially newer versions.

flowln:
> This kind of thing does not generally require any action on our part. Package metadata on each release of datamodel-code-generator should indicate which versions of Python it supports, and pip should get the most recent release that supports the appropriate Python version. So, in a Python 3.8 environment, we would get the latest release that supported 3.8, and in a Python 3.9+ environment, we would get potentially newer versions.

The issue here is that the schema generator code here in this code base hardcodes the 3.8 version: https://github.com/bluesky/event-model/blob/7cb9543ecbec1c3d87328ef087276c8b522a0ee0/src/event_model/generate/create_documents.py#L118

This in theory means that `event-model` itself should only support Python 3.8, which is definitely not ideal. I'm not sure why this was made like that, as the commit message introducing it does not explain it, and I really don't know what's the consequence of changing this version, so I opted to restrict the dependency version to be the least intrusive possible while still generating a valid dev environment.

evalott100:
>  I'm not sure why this was made like that, as the commit message introducing it does not explain it

Apologies, when I made this PR [bluesky was still supporting 3.8](https://github.com/bluesky/bluesky/blob/7c2a44f09a747f59a17abc6cfdfe934c61c266fb/pyproject.toml#L10), and as such I wanted the generated `TypedDict` to be compatible with it.

We've removed 3.8 from supported versions of bluesky so I'm happy for this to be bumped to

``` python
        target_python_version=datamodel_code_generator.PythonVersion.PY_39,
```

pyproject.toml
```diff
@@ -51,7 +51,7 @@ dev = [
 
   # For schema generation.
   "pydantic<3",
-  "datamodel-code-generator",
+  "datamodel-code-generator<0.28",
```
evalott100, line 54:
```suggestion
  "datamodel-code-generator",
```

src/event_model/basemodels/event_descriptor.py
```diff
@@ -193,7 +193,7 @@ class DataKey(BaseModel):
         str, Field(description="The source (ex piece of hardware) of the data.")
     ]
     units: Annotated[
-        str,
+        Optional[str],
         Field(default="", description="Engineering units of the value"),
```
evalott100, line 197: We probably want to make the default in the basemodels `None` if that's what bluesky is doing.

```suggestion
        Field(default=None, description="Engineering units of the value"),
```

evalott100 (changes requested):
Thanks very much!

I think we should just remove `target_python_version=` as an argument, it'll end up being more trouble than it's worth. For now there is no difference in generated schema.

evalott100 (approved):
Thanks!

Issue #351, opened 2025-04-14
Express ragged arrays in Event Descriptor `shape`


danielballan:
Let us relax the schema for `shape` to accept `null`/`None` for a dimension to express "variable size". See https://github.com/bluesky/tiled/issues/801 for motivation.

https://github.com/bluesky/event-model/blob/33cdfaf72f6d791d719b4923dbd0632a48817b7f/src/event_model/schemas/event_descriptor.json#L123-L130

PR #352, opened 2025-04-14
Allow None in descriptor shape

Tom-Willemsen:
## Description

Closes https://github.com/bluesky/event-model/issues/351

Allow `None` as a shape in `event-model` - a dimension with extent `None` is interpreted as a dimension of unknown shape.

## Motivation and Context

Fixes https://github.com/bluesky/event-model/issues/351

One scientific motivation for this is event-mode data, common at neutron facilities, where a detector records the position and timestamp of each detection event individually. This means that event-mode detectors conceptually give an array of variable shape each time data is acquired from them, where the dimensionality corresponds to the number of detection events.

## How Has This Been Tested?

Tested that the `RunEngine` now _accepts_ emitting descriptors with a declared shape of `(None, )`.

At the moment not all downstream consumers are able to process these events - further work will be required to support this. For example the `TiledWriter` in bluesky does not currently handle this.

This change should be backwards-compatible with all existing data, as it is a loosening of the schema.

### DIFF
--- a/src/event_model/basemodels/event_descriptor.py
+++ b/src/event_model/basemodels/event_descriptor.py
@@ -186,8 +186,11 @@ class DataKey(BaseModel):
         ),
     ]
     shape: Annotated[
-        List[int],
-        Field(description="The shape of the data.  Empty list indicates scalar data."),
+        List[Optional[int]],
+        Field(
+            description="The shape of the data.  Empty list indicates scalar data. "
+            "None indicates a dimension with unknown length."
+        ),
     ]
     source: Annotated[
         str, Field(description="The source (ex piece of hardware) of the data.")
--- a/src/event_model/documents/event_descriptor.py
+++ b/src/event_model/documents/event_descriptor.py
@@ -127,9 +127,9 @@ class DataKey(TypedDict):
     """
     Number of digits after decimal place if a floating point number
     """
-    shape: List[int]
+    shape: List[Optional[int]]
     """
-    The shape of the data.  Empty list indicates scalar data.
+    The shape of the data.  Empty list indicates scalar data. None indicates a dimension with unknown length.
     """
     source: str
     """
--- a/src/event_model/schemas/event_descriptor.json
+++ b/src/event_model/schemas/event_descriptor.json
@@ -122,10 +122,17 @@
                 },
                 "shape": {
                     "title": "shape",
-                    "description": "The shape of the data.  Empty list indicates scalar data.",
+                    "description": "The shape of the data.  Empty list indicates scalar data. None indicates a dimension with unknown length.",
                     "type": "array",
                     "items": {
-                        "type": "integer"
+                        "anyOf": [
+                            {
+                                "type": "integer"
+                            },
+                            {
+                                "type": "null"
+                            }
+                        ]
                     }
                 },
                 "source": {


src/event_model/documents/event_descriptor.py
```diff
@@ -127,9 +127,9 @@ class DataKey(TypedDict):
     """
     Number of digits after decimal place if a floating point number
     """
-    shape: List[int]
+    shape: List[Optional[int]]
     """
-    The shape of the data.  Empty list indicates scalar data.
+    The shape of the data.  Empty list indicates scalar data. None indicates a dimension with unknown length.
```
danielballan, line 132:
```suggestion
    The shape of the data.  Empty list indicates scalar data. None indicates a dimension with unknown or variable length.
```

src/event_model/schemas/event_descriptor.json
```diff
@@ -122,10 +122,17 @@
                 },
                 "shape": {
                     "title": "shape",
-                    "description": "The shape of the data.  Empty list indicates scalar data.",
+                    "description": "The shape of the data.  Empty list indicates scalar data. None indicates a dimension with unknown length.",
```
danielballan, line 125:
```suggestion
                    "description": "The shape of the data.  Empty list indicates scalar data. None indicates a dimension with unknown or variable length.",
```

src/event_model/basemodels/event_descriptor.py
```diff
@@ -186,8 +186,11 @@ class DataKey(BaseModel):
         ),
     ]
     shape: Annotated[
-        List[int],
-        Field(description="The shape of the data.  Empty list indicates scalar data."),
+        List[Optional[int]],
+        Field(
+            description="The shape of the data.  Empty list indicates scalar data. "
+            "None indicates a dimension with unknown length."
```
danielballan, line 192:
```suggestion
            "None indicates a dimension with unknown or variable length."
```

danielballan (commented):
Thanks @Tom-Willemsen! One minor suggestion.

evvaaaa:
Nice change, lgtm

evvaaaa (approved):

jacopoabramo:
Just a little note since this change is also useful for me and this was discussed in the last dev meeting:

- it would be nice (if possible) in the future to enhance the schema with another type of document which can attest the change of a descriptor during a run (for example ChangeEvent or something like that); 
- a use case I can attest for this is that during a run slow changing properties such as the region of interest may be changed while the run is being executed (for example while capturing live from a detector);
  - using the "configure" command would invalidate the descriptor document previously produced, as stated [here](https://github.com/bluesky/bluesky/pull/1888#issuecomment-2678604380)

At run engine level I don't precisely know what this would imply (although I feel it would require big effort), but I suspect it's too big of a feature to implement on the fly and right now for what I need this PR is more than enough (hopefully). Just wanted to mention it for the future.

danielballan (approved):

DiamondJoseph:
As down stream consumers I'd prefer `Optional[List[int]]` to `List[Optional[int]]`: is it one dimension in a list of other known dimensions that will be None or is it either you get the shape or you don't?

Can unknown shape be coerced into an ophyd-async `Stream` document?

danielballan:
Good question, this is worth examining.

`List[Optional[int]]` conveys more information than `Optional[List[int]]`. For example, `[None, None, None]` at least tells the consumer that they can expect a three-dimensional array. We have some use cases where outer dimensions are known and inner vary, as in `[3, 5, None]` or `[3, None, None]`. It is useful to convey the fixed dimensions.

Also, `List[Optional[int]]` complies with the [Array API definition of `array.shape`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.shape.html#array_api.array.shape). To be precise, the Array API's intention is to use `None` to indicate _unknown_ size, and here we are stretching that a bit to mean _unknown or variable_ size, following in the footsteps of the [ragged](https://github.com/scikit-hep/ragged?tab=readme-ov-file) library.

More background is covered in https://github.com/bluesky/tiled/issues/801. Tiled does not yet handle ragged shapes, but we have described how it could be added.

Consumers will need to be updated to either ignore ragged data; raise; pad to the largest size with zero, NaN, or another fill value; truncate to the smallest size; or handle it properly (if possible). HDF5 has no native support for ragged structures, so padding seems like a likely path there.

Tom-Willemsen:
> Can unknown shape be coerced into an ophyd-async Stream document?

Is the question about [this](https://github.com/bluesky/ophyd-async/blob/48236a919cec15700bbeeda2324ab2efff1c4558/src/ophyd_async/core/_hdf_dataset.py#L26) or have I misunderstood? If so, yes, a similar PR will be needed on ophyd-async to match event-model's definitions - which I was planning to make after/if this one gets through and we've ironed out `Optional[list]` vs `list[Optional]` here.

For my initial use-case, the data is 1-dimensional so `None` vs `[None]` makes little difference in that case. The only difference is that with `[None]`, bluesky tells me it's 1D, whereas with `None` I have to "know" that from some external context. I also like the argument above about being able to describe arrays with partially known dimensionality - I can imagine data structures where that may be useful...

evvaaaa:
> ```shape=None or [] #scalar value```

I would vote for 

* `None` or `NotRequired` for a scalar.
* `[]` for an array of unknown dimension.
* `[None, None, <etc...>]` for an array of unknown exact shape, but known dimension.

danielballan:
We have always required the `shape` key. For scalar, I would favor sticking with `None`, not loosening the specificiation to accept `NotRequired` as a synonym.

Does anyone have a use case for an array of unknown _dimension_? That is hard for consumers to handle, so I would resist supporting it unless we have a compelling need. For example, HDF5 can handle a dimension of unknown/variable length but not an unknown dimensionality. Likewise, Tiled can grow the ability to represent the former, but I have no idea how I would implement the latter.

Where use cases do arise, I think we can still say something more than "unknown dimensionality". For example, imagine a detector that might produce a single image or a stream of images: this could be represented like `[None, 512, 512]` where the first dimension may be `1` if we receive a single image. In the very worst case, we can write `[None, None, <etc>]` to express the maximum number of dimensions that might possible arise in the data.

coretl:
> Does anyone have a use case for an array of unknown _dimension_? That is hard for consumers to handle, so I would resist supporting it unless we have a compelling need. For example, HDF5 can handle a dimension of unknown/variable length but not an unknown dimensionality. Likewise, Tiled can grow the ability to represent the former, but I have no idea how I would implement the latter.

I can't think of a use case for it, so I would say we don't support it.

To paste the table we made in a previous collab call:
Measurement | Descriptor | HDF mimetype | Tiled
-- | -- | -- | --
Always scalar | `[]` or `None` | `[N]` | `[N]`
Can give M scalars per Event | `[M]`, where `M` may be `1` | `[M*N]` | `[N, M]`
Variable number of scalars per Event | `[None]` | `[?]` | `[N, None]`
Single `M`-length spectra per Event | `[1, M]` | `[N, M]` | `[N, 1, M]`
`Q` `M`-length spectra per Event | `[Q, M]` | `[N*Q, M]` |`[N, Q, M]`
Area Detector taking `M` 2D images | `[M, Y, X]` | `[M*N, Y, X]` | `[N, M, Y, X]`

I don't think we have changed anything here, although I would suggest we drop `None` as a possible shape, and require an explicit `[]` for the "Always scalar" case

PR #353, opened 2025-04-15
Pin pydantic to <2.11, newer versions not supported on Python 3.8

Tom-Willemsen:
## Description

Pin `pydantic<2.11` (dev dependency only)

Pydantic 2.11 is not supported on Python 3.8, which leads to different schema generation on 3.8 versus later versions. Force all the python versions to pick up the same pydantic version.

The other option would be to drop Python 3.8 support in this repo - bluesky itself no longer supports 3.8, do other uses of `event-model` still need it?

## Motivation and Context

At the moment it is not possible to create a PR which passes both the 3.8 build and >3.8 builds

## How Has This Been Tested?

CI builds

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ dev = [
   "numpydoc",
 
   # For schema generation.
-  "pydantic<3",
+  "pydantic<2.11",
   "datamodel-code-generator",
 ]
 


DiamondJoseph:
@evvaaaa is there a timeline for deprecating 3.8?

evvaaaa:
> is there a timeline for deprecating 3.8?

I'd be happy to drop it [now that bluesky has](https://github.com/bluesky/bluesky/releases/tag/v1.13.1). 

I've been quite busy so I haven't got around to making a new release in a while. I think we should drop 3.8 and then do it.

prjemian:
Thought we (blueskyproject "We") agreed to only support the official list of supported Python versions. That means 3.8 can be dropped now and 3.9 can be dropped after October 1.

prjemian:
The conversation considered a more aggressive schedule but settled on this. Should verify since this is based on my memory.

PR #364, opened 2025-08-26
ENH: generate `Document` union type

jacopoabramo:
## Description

Add a `Document` union type; differently from `DocumentType`, this type union is for type checking instances of document objects rather than their class type.

I simply reshuffled a bit the `GENERATED_INIT_PY` template to keep the original `DocumentType` intact.

## Motivation and Context

I was trying to type hinting a callback function that should accept all possible documents and realized that `DocumentType` wasn't fit for the purpose.

## How Has This Been Tested?

Tested locally via `pytest`.

### DIFF
--- a/src/event_model/basemodels/__init__.py
+++ b/src/event_model/basemodels/__init__.py
@@ -29,6 +29,19 @@ DocumentType = Union[
     Type[StreamResource],
 ]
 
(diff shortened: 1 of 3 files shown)


evvaaaa (approved):
LGTM!

PR #365, opened 2025-08-27
Add Document as a Union of all document types

tpoliaw:


tpoliaw:
CI is failing with `ALL_BASEMODELS` not being a valid arg to `Union` but python is happy to accept a tuple of types, eg `Union[int, str] == Union[(int, str)]` is `True`. Can duplicate the list for `Document` or add a `type: ignore` comment.

evvaaaa:
> Can duplicate the list for Document

That's what #364 is doing (should be ready to merge now).

Probably best to duplicate rather than add a `# type: ignore`.

DiamondJoseph:
Closing as duplicate of #364 
@evvaaaa can we get a release of event-model?

evvaaaa:
Will add some release notes now.

#366

Issue #354, opened 2025-04-15
Unpin pydantic once 3.8 is unsupported


evvaaaa:
In https://github.com/bluesky/event-model/pull/353, we pinned pydantic because the latest release doesn't work with python 3.8.

We should unpin it once we drop 3.8 support.

jacopoabramo:
@evvaaaa #368 addressed this so the issue can be closed (I forgot to edit the message correctly...)

PR #368, opened 2025-10-06
ENH: drop python 3.8, 3.9 and support 3.13

### LINKED ISSUE #367: Drop Python 3.8/3.9 and add support for 3.13
Python 3.9 will reach EOL end of this month.

This [PR in the bluesky repo](https://github.com/bluesky/bluesky/pull/1959) is moving towards ending support for 3.9 and adding support for 3.13.

Could the same be applied for the event model?

jacopoabramo:
## Description
Closes #367 and #354 

## Motivation and Context

Python 3.9 reaches EOL end of this month; this is an attempt to mirror bluesky repo.

~Currently blocked by [#1959](https://github.com/bluesky/bluesky/pull/1959)~

## How Has This Been Tested?

Tested locally but some changes due to autogenerating typed dictionary models are fishy.

### DIFF
--- a/src/event_model/basemodels/__init__.py
+++ b/src/event_model/basemodels/__init__.py
@@ -1,5 +1,3 @@
-from typing import Tuple, Type, Union
-
 from event_model.basemodels.datum import Datum
 from event_model.basemodels.datum_page import DatumPage
 from event_model.basemodels.event import Event
@@ -16,33 +14,33 @@ from event_model.basemodels.run_stop import RunStop
 from event_model.basemodels.stream_datum import StreamDatum
 from event_model.basemodels.stream_resource import StreamResource
 
-DocumentType = Union[
-    Type[Datum],
-    Type[DatumPage],
-    Type[Event],
-    Type[EventDescriptor],
-    Type[EventPage],
-    Type[Resource],
-    Type[RunStart],
-    Type[RunStop],
-    Type[StreamDatum],
-    Type[StreamResource],
-]
+DocumentType = (
+    type[Datum]
+    | type[DatumPage]
+    | type[Event]
+    | type[EventDescriptor]
+    | type[EventPage]
+    | type[Resource]
+    | type[RunStart]
+    | type[RunStop]
+    | type[StreamDatum]
+    | type[StreamResource]
+)
 
-Document = Union[
-    Datum,
-    DatumPage,
-    Event,
-    EventDescriptor,
-    EventPage,
-    Resource,
-    RunStart,
-    RunStop,
-    StreamDatum,
-    StreamResource,
-]
+Document = (
+    Datum
+    | DatumPage
+    | Event
+    | EventDescriptor
+    | EventPage
+    | Resource
+    | RunStart
+    | RunStop
+    | StreamDatum
+    | StreamResource
+)
 
-ALL_BASEMODELS: Tuple[DocumentType, ...] = (
+ALL_BASEMODELS: tuple[DocumentType, ...] = (
     Datum,
     DatumPage,
     Event,
--- a/src/event_model/documents/run_start.py
+++ b/src/event_model/documents/run_start.py
@@ -4,18 +4,29 @@
 
 from __future__ import annotations
 
-from typing import Any, Dict, List, TypedDict, Union
+from typing import Any, Literal, TypeAlias, TypedDict
 
 from typing_extensions import NotRequired
 
 
+class Hints(TypedDict):
+    """
+    Start-level hints
+    """
+
+    dimensions: NotRequired[list[list[list[str] | str]]]
+    """
+    The independent axes of the experiment. Ordered slow to fast
+    """
+
+
 class Calculation(TypedDict):
-    args: NotRequired[List]
+    args: NotRequired[list[Any]]
     callable: str
     """
     callable function to perform calculation
     """
-    kwargs: NotRequired[Dict[str, Any]]
+    kwargs: NotRequired[dict[str, Any]]
     """
     kwargs for calcalation callable
     """
@@ -25,46 +36,35 @@ class ConfigurationProjection(TypedDict):
     config_device: str
     config_index: int
     field: str
-    location: NotRequired[str]
+    location: Literal["configuration"]
     """
     Projection comes from configuration fields in the event_descriptor document
     """
     stream: str
-    type: NotRequired[str]
+    type: Literal["linked"]
     """
     Projection is of type linked, a value linked from the data set.
     """
 
 
-DataType = Any
-
-
-class Hints(TypedDict):
-    """
-    Start-level hints
-    """
-
-    dimensions: NotRequired[List[List[Union[List[str], str]]]]
-    """
-    The independent axes of the experiment. Ordered slow to fast
-    """
+DataType: TypeAlias = dict[str, "DataType"]
 
 
 class LinkedEventProjection(TypedDict):
     field: str
-    location: NotRequired[str]
+    location: Literal["event"]
     """
     Projection comes and event
     """
     stream: str
-    type: NotRequired[str]
+    type: Literal["linked"]
     """
     Projection is of type linked, a value linked from the data set.
     """
 
 
 class StaticProjection(TypedDict):
-    type: NotRequired[str]
+    type: Literal["static"]
     """
     Projection is of type static, a value defined here in the projection
     """
@@ -80,12 +80,12 @@ class CalculatedEventProjection(TypedDict):
     required fields if type is calculated
     """
     field: str
-    location: NotRequired[str]
+    location: Literal["event"]
     """
     Projection comes and event
     """
     stream: str
-    type: NotRequired[str]
+    type: Literal["calculated"]
     """
     Projection is of type calculated, a value that requires calculation.
     """
@@ -96,7 +96,7 @@ class Projections(TypedDict):
     Describe how to interperet this run as the given projection
     """
 
-    configuration: Dict[str, Any]
+    configuration: dict[str, Any]
     """
     Static information about projection
     """
@@ -104,14 +104,12 @@ class Projections(TypedDict):
     """
     The name of the projection
     """
-    projection: Dict[
+    projection: dict[
         str,
-        Union[
-            ConfigurationProjection,
-            LinkedEventProjection,
-            CalculatedEventProjection,
-            StaticProjection,
-        ],
+        ConfigurationProjection
+        | LinkedEventProjection
+        | CalculatedEventProjection
+        | StaticProjection,
     ]
     version: str
     """
@@ -125,7 +123,7 @@ class RunStart(TypedDict):
     later documents link to it
     """
 
-    data_groups: NotRequired[List[str]]
+    data_groups: NotRequired[list[str]]
     """
     An optional list of data access groups that have meaning to some external system. Examples might include facility, beamline, end stations, proposal, safety form.
     """
@@ -147,8 +145,8 @@ class RunStart(TypedDict):
     """
     Name of project that this run is part of
     """
-    projections: NotRequired[List[Projections]]
-    sample: NotRequired[Union[Dict[str, Any], str]]
+    projections: NotRequired[list[Projections]]
+    sample: NotRequired[dict[str, Any] | str]
     """
     Information about the sample, may be a UID to another collection
     """
--- a/src/event_model/generate/create_documents.py
+++ b/src/event_model/generate/create_documents.py
@@ -1,7 +1,8 @@
 import json
+import warnings
 from collections import OrderedDict
 from pathlib import Path
-from typing import Any, Dict, Type, cast
+from typing import Any, cast
 
 import datamodel_code_generator
 from pydantic import BaseModel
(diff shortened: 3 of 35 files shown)


src/event_model/basemodels/__init__.py
```diff
@@ -16,20 +16,20 @@
...
-    Type[DatumPage],
-    Type[Event],
-    Type[EventDescriptor],
-    Type[EventPage],
-    Type[Resource],
-    Type[RunStart],
-    Type[RunStop],
-    Type[StreamDatum],
-    Type[StreamResource],
+DocumentType = Union[  # noqa: UP007
```
jacopoabramo: I figured it would be easier to just tell ruff to shut up than to reformat everything in a single line...

src/event_model/documents/run_start.py
```diff
@@ -25,12 +25,12 @@ class ConfigurationProjection(TypedDict):
     config_device: str
     config_index: int
     field: str
-    location: NotRequired[str]
+    location: Literal["configuration"]
```
jacopoabramo, line 39: I'm not exactly sure what happened here; from the looks of it the JSON documents aren't changed (I would have seen it after launching the regeneration) and yet datamodel changed these `NotRequired` fields to `Literal`...

It happens also in other places

src/event_model/generate/create_documents.py
```diff
@@ -115,6 +115,7 @@ def generate_typeddict(jsonschema_path: Path, documents_root=TYPEDDICTS) -> Path
         input_file_type=datamodel_code_generator.InputFileType.JsonSchema,
         output=document_path,
         output_model_type=datamodel_code_generator.DataModelType.TypingTypedDict,
+        target_python_version=datamodel_code_generator.PythonVersion.PY_310,
```
jacopoabramo: This is the only thing I changed when calling datamodel, the rest is just formatting mostly

jacopoabramo:
@evvaaaa any suggestions? Apparently some fields have been changed from `NotRequired` to `Literal` during the generation process (after setting the minimum python version to 3.10).

src/event_model/basemodels/__init__.py
evvaaaa: Is `type[Datum] | type[DatumPage] | ...` not supported from 3.10 on?

src/event_model/documents/run_start.py
evvaaaa: It is more true to the pydantic model so it's better. I don't exactly know why this would change, maybe because we're specifying 3.10 as our target, if this also happens in main then it's possible it's just datamodel supporting literals now.

Either way, it's good for us, it's more accurate.

https://github.com/bluesky/event-model/blob/a896eb2145359f2b96ceca6108a143340f731af1/src/event_model/basemodels/run_start.py#L64

evvaaaa (changes requested):
This would be nice, however, bluesky supports 3.9 still, and event-model is a required dependency. I wouldn't be happy updating this without also dropping 3.9 from bluesky.

jacopoabramo:
> This would be nice, however, bluesky supports 3.9 still

There's a [PR open for that](https://github.com/bluesky/bluesky/pull/1959). Once that's merged, this can be merged as well safely I believe

src/event_model/basemodels/__init__.py
jacopoabramo: It is, but the generation process would force to put everything on a single line as this type of syntax:

```py
MyInt = type[int] |
type[str] |
type[bytes] |
```

doesn't seem to be supported.

I *could* do that on a single line but I feel ruff would complain.

src/event_model/documents/run_start.py
jacopoabramo: I'll check the tests and update them then.

src/event_model/basemodels/__init__.py
jacopoabramo: Ah, nevermind:

```py
MyInt = (
    type[int] |
    type[str] |
    type[bytes]
)
```

this seems to be a valid syntax and it is interpreted correctly as the union of multiple types. I'll apply changes.

src/event_model/documents/run_start.py
jacopoabramo: So apparently the tests were fine, for some reasons the JSON files weren't correctly updated. Now they are.

evvaaaa (approved):
To be merged alongside the bluesky update.

jacopoabramo:
Happy new year @evvaaaa ; just a friendly ping that [#1959](https://github.com/bluesky/bluesky/pull/1959) has been merged.

evvaaaa:
> Happy new year @evvaaaa ; just a friendly ping that https://github.com/bluesky/bluesky/pull/1959 has been merged.

Happy new year to you too :partying_face: 

Cool, happy for this to be merged. Can you rebase and squash it?

jacopoabramo:
@evvaaaa it seems the latest release of datamodel-code-generator is switching entirely to ruff for formatting; is it ok for me to adapt that change to this PR?

EDIT: according to the error message the warning can be suppressed by explicitly setting the formatters in the generate function. Since event-model is already using ruff, I can simply add it in the API call

jacopoabramo:
Ah, nevermind, the test actually fails because `PydanticDeprecatedSince20` apparently is not raised... but we could it two birds with one stone

evvaaaa:
Some of our changes to datamodel-code-generator have also been merged. I can make an issue, will let us cut down on custom jinja.

For now we just fix it by doing what they [suggest](https://github.com/bluesky/event-model/actions/runs/20750423334/job/59578078960#step:7:224).

jacopoabramo:
@evvaaaa I applied their suggestion locally; trouble is that it produces a series of exceptions. Copy-pasting the local output of my `tox -e tests` command:

```
  + Exception Group Traceback (most recent call last):
  |   File "C:\git\event-model\.venv\lib\site-packages\_pytest\runner.py", line 353, in from_call
  |     result: TResult | None = func()
  |   File "C:\git\event-model\.venv\lib\site-packages\_pytest\runner.py", line 245, in <lambda>
  |     lambda: runtest_hook(item=item, **kwds),
  |   File "C:\git\event-model\.venv\lib\site-packages\pluggy\_hooks.py", line 512, in __call__
  |     return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
  |   File "C:\git\event-model\.venv\lib\site-packages\pluggy\_manager.py", line 120, in _hookexec
  |     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
  |   File "C:\git\event-model\.venv\lib\site-packages\pluggy\_callers.py", line 167, in _multicall
  |     raise exception
  |   File "C:\git\event-model\.venv\lib\site-packages\pluggy\_callers.py", line 139, in _multicall
  |     teardown.throw(exception)
  |   File "C:\git\event-model\.venv\lib\site-packages\_pytest\logging.py", line 858, in pytest_runtest_teardown
  |     yield
  |   File "C:\git\event-model\.venv\lib\site-packages\pluggy\_callers.py", line 139, in _multicall
  |     teardown.throw(exception)
  |   File "C:\git\event-model\.venv\lib\site-packages\_pytest\capture.py", line 905, in pytest_runtest_teardown
  |     return (yield)
  |   File "C:\git\event-model\.venv\lib\site-packages\pluggy\_callers.py", line 121, in _multicall
  |     res = hook_impl.function(*args)
  |   File "C:\git\event-model\.venv\lib\site-packages\_pytest\unraisableexception.py", line 163, in pytest_runtest_teardown
  |     collect_unraisable(item.config)
  |   File "C:\git\event-model\.venv\lib\site-packages\_pytest\unraisableexception.py", line 81, in collect_unraisable
  |     raise ExceptionGroup("multiple unraisable exception warnings", errors)
  | exceptiongroup.ExceptionGroup: multiple unraisable exception warnings (10 sub-exceptions)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "C:\git\event-model\.venv\lib\site-packages\datamodel_code_generator\format.py", line 347, in format_code
    |     code = self.apply_ruff_check_and_format(code)
    | ResourceWarning: unclosed file <_io.BufferedReader name=17>
    |
    | The above exception was the direct cause of the following exception:
    |
    | Traceback (most recent call last):
    |   File "C:\git\event-model\.venv\lib\site-packages\_pytest\unraisableexception.py", line 67, in collect_unraisable
    |     warnings.warn(pytest.PytestUnraisableExceptionWarning(msg))
    | pytest.PytestUnraisableExceptionWarning: Exception ignored in: <_io.FileIO name=17 mode='rb' closefd=True>
    | Enable tracemalloc to get traceback where the object was allocated.
    | See https://docs.pytest.org/en/stable/how-to/capture-warnings.html#resource-warnings for more info.
 ```
 
The specific exception of the exception group is raised 10 times in total (a.k.a. for each JSON schema document currently generated, except for the `bulk_` documents which are deprecated anyway). From the looks of it, it tries to run `ruff check/format` as a subprocess on an currently opened file and it can't apply modifications. I'm not sure if this is a bug on `datamodel` or in the our generation script; to keep things simple I would suggest to filter out the `FutureWarning` produced and notify this behavior. Since it's not inherent to the generation process and we run `ruff` regardless of the outcome it should not cause issues.

**Comment by jacopoabramo** [comment]
I suppressed the warning on the `generate` function, so that it does not propagate when running `python -m event_model.generate`; I understand it's not the optimal choice but I would rather get this PR out and try to fix the warning later on.

**Comment by jacopoabramo** [comment]
The failing type checking test is caused by the fact that the latest version of `datamodel-code-generator` is issuing the `closed` parameter to initialize the typed dictionary subclasses to enforce the rule that documents cannot be extended with additional keys ([PEP 728](https://peps.python.org/pep-0728/)). Apparently mypy still [doesn't support this feature](https://github.com/python/mypy/issues/18176).

What's your reccomendation? Can this be added as an ignore for mypy?

Although the more I progress with this PR the more troublesome it becomes...

EDIT: maybe it could be solved via editing the ninja template

**Comment by jacopoabramo** [comment]
I opened [#2951](https://github.com/koxudaxi/datamodel-code-generator/issues/2951) to see how to suppress this. Most probably it will involve adapting the template.

**Comment by jacopoabramo** [comment]
@evvaaaa I can vendor the original [`TypedDict` jinja template](https://github.com/koxudaxi/datamodel-code-generator/blob/main/src/datamodel_code_generator/model/template/TypedDictClass.jinja2) from `datamodel`. I would still prefer the CLI option nevertheless; the only concern I have is that the template has to pin-point typed dictionaries that have no inheritance, since this `closed` parameter is not generated for typed dictionaries that may be used as a baseline to extend a document.

I get the impression that `closed=True` might be caused by the presence of `"additionalProperties": false` in the specific schemas (i.e. the conversion from `run_start.json` produces a `closed` typed dictionary, while `event.json` does not and at first glance it's their main difference in terms of structure).

**Comment by jacopoabramo** [comment]
@evvaaaa [#2951](https://github.com/koxudaxi/datamodel-code-generator/issues/2951) has been resolved with `datamodel-code-generator>=0.53.0`; I removed the template directory and pinned the version of the code generator to stay on the safe side. Feel free to review. The only nitpick is the annoying ignore filter in the `generate` call but other than that everything checks out.

File: src/event_model/generate/create_documents.py [hunk 4]
```diff
@@ -242,13 +244,20 @@ def generate_init_py(output_root: Path):
 
     document_types = "\n".join(
         [
-            f"    Type[{class_name}],  # noqa: F405,"
-            for class_name in document_class_names
+            f"    type[{class_name}]  # noqa: F405,"
+            if i == 0
+            else f"    | type[{class_name}]  # noqa: F405"
+            for i, class_name in enumerate(document_class_names)
```
evvaaaa: NIT:

Easier for me to read

```suggestion
document_types = "    " + "\n    | ".join([
        f"type[{class_name}]  # noqa: F405"
        for class_name in document_class_names
])
```

src/event_model/generate/create_documents.py
```diff
@@ -242,13 +244,20 @@ def generate_init_py(output_root: Path):
...
     )
 
     documents = "\n".join(
-        [f"    {class_name},  # noqa: F405" for class_name in document_class_names]
+        [
+            f"    {class_name}  # noqa: F405"
+            if i == 0
+            else f"    | {class_name}  # noqa: F405"
+            for i, class_name in enumerate(document_class_names)
+        ]
```
evvaaaa: ditto

src/event_model/basemodels/__init__.py
evvaaaa: We should change this to `|` instead of `typing.Union`, same with `Document` below.

evvaaaa (changes requested):
Thanks for this, turned into a much larger PR than anticipated :)

src/event_model/generate/create_documents.py
jacopoabramo: Done for both

src/event_model/basemodels/__init__.py
jacopoabramo: Done

evvaaaa (approved):

evvaaaa:
Maybe do the rebase an push here before merging, I'm not sure what is out of sync :thinking:

jacopoabramo:
Done; last update was 5 months ago anyway... :U

evvaaaa:
Will do a minor release as soon as bluesky does.

PR #373, opened 2026-04-29
Fix Memory Leak in RunRouter

genematx:
Fix memory leak: clean up _resources and _stream_resources in RunRouter.stop()

## Motivation and Context

`_resources` was keyed by `resource_uid` but `stop()` was popping with `start_uid`, so every Resource document from every completed run accumulated in `_resources `indefinitely (the pop was always a no-op).

`_stream_resources` was never cleaned up in `stop()` at all.

## Description

Fix both by introducing `_start_to_resources` and `_start_to_stream_resources` reverse-lookup maps (mirroring the existing `_start_to_descriptors` pattern). These are populated in `resource()` and `stream_resource()` and used in `stop()` to efficiently remove all per-run entries without an O(N) scan.

## How Has This Been Tested?

Adds a regression test covering both code paths across multiple runs.

### DIFF
--- a/src/event_model/__init__.py
+++ b/src/event_model/__init__.py
@@ -1529,6 +1529,11 @@ class RunRouter(DocumentRouter):
         self._resources: dict = {}
         self._stream_resources: dict = {}
 
+        # Map RunStart UID to the list of Resource UIDs and StreamResource UIDs.
+        # These are used to facilitate efficient cleanup of the caches above.
+        self._start_to_resources: defaultdict = defaultdict(list)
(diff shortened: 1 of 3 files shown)


danielballan:
Old resource documents do not have a `run_start` reference (under any name). I believe we intentionally retain/leak Resources for this reason.

If RunRouter is used to migrate old Resource documents, this issue is still relevant today. Perhaps we could put it behind a feature flag that does _not_ leak by default, but _can_ leak when we need to push very old Resource documents through a RunRouter.

genematx:
Hmmm, interesting. thanks Dan

PR #375, opened 2026-05-26
Added object classes to descriptor

evvaaaa:
Added `object_classes` key to the descriptor schema and compose method.

Event-model component of https://github.com/bluesky/bluesky/issues/2013

### DIFF
--- a/src/event_model/__init__.py
+++ b/src/event_model/__init__.py
@@ -2445,6 +2445,7 @@ class ComposeDescriptor:
         hints=None,
         configuration=None,
         object_keys=None,
+        object_classes=None,
         time=None,
         uid=None,
         validate=True,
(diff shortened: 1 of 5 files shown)


evvaaaa:
@danielballan 

This new key is not required in the compose method and schema, so we can merge it and minor release event model now, latest release bluesky will still function with this event model, though CI will fail on bluesky `main` because of the presence of new keys.

danielballan (approved):
Sorry for the delay, just catching up on GH notifications today.

Which part of the bluesky CI do you expect to fail? I'm sure you're right but I'd like to have clearly in my mind which bluesky tests on `main` merging and releasing this will break.

Otherwise, happy to move forward with this.

danielballan:
Or do you just mean that https://github.com/bluesky/bluesky/pull/2024 will fail _until_ we merge and release this?

evvaaaa:
> Or do you just mean that https://github.com/bluesky/bluesky/pull/2024 will fail until we merge and release this?

I had a weird environment and `DRAFT_0_STREAM_RESOURCE` was enabled because of the added commit rev. Can confirm everything passes (locally) on bluesky `main` with this branch. I will merge it and make release notes.

PR #378, opened 2026-07-26
ENH: add named docs tuple

### LINKED ISSUE #369: Create a union of all supported document names and documents to allow type inference
Callbacks to the bluesky RE are currently called as `callback(name, doc)` which does not allow inferring the structure of doc when given a particular name. If we passed a tuple of the two and called `callback(named_doc)` then we could do type inference on the content. E.g.

Code sample in [pyright playground](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgCYAiSAxjADRQAySMxIAhgDb0CqKCbVANYAoYVQ5sAzpKjkwVAIIAKIgjKUaASgBcwqPqjAwYbVEkwQegwCM2IU6hijxUmXKoAhFSQrUYOq30jEzMLQKhbGFNgDjA2J2EAOTYIdXkAV1SUfABeKAAdcJh0hA5iAG0mFnYOcoAiUnkAfTY6gF16dwU2qAAfAqKSssrmVk56xqom63bO%2BQ820WFSYmAoCDgmqk4OWyElFBSyJsnM4mzTZNSKDKz-XQN1%2BKoACyhD65Pb86jwx%2B3JMQoA1mq16JMHo8ofoUJkoHlJvVbCB2n8DACgSCpjNwfJIdDHrDoAj5Ej4qjHggQI4DplNKIgA)

```python
from typing import TypedDict, Literal, Unpack

class DocA(TypedDict):
    foo: str
    bar: int

class DocB(TypedDict):
    foo: str
    bat: float

NamedDocument = \
    tuple[Literal["doc_a"], DocA] | \
    tuple[Literal["doc_b"], DocB]


def my_callback(named_document: NamedDocument):
    match named_document:
        case "doc_a", doc:
            num = doc["bar"]
        case "doc_b", doc:
            num = doc["bat"]
    print(num)


```

This ticket is to decide if this is a good idea, then create the `NamedDocument` union so that a corresponding bluesky ticket can be created to discuss the callback changes on the bluesky time.

## Acceptance Criteria
- Decide this is a good idea, then create the union

jacopoabramo:
Closes #369

I got some help from Claude for this. I honestly let him roll the tests on its own to give it a whirl but have NOT properly reviewed them. I'm opening the PR but I'll leave it as draft to confirm that the tests make sense, hoping to get another pair of eyes as well.

The PR does two things:

- vendors [`backports.strenum`](https://pypi.org/project/backports.strenum/) to convert `DocumentNames` to a `StrEnum`; when Python 3.10 is dropped from support the vendor can be removed;
- adds a `DocWrapper` helper for leveraging match-case on the document name.

`DocWrapper` is a named tuple that encapsulates the document name as an enum value from DocumentNames and the actual document type. Typically callbacks are type annotated as:

```py
# Document is the union of all possible
# document types
from event_model import Document

def my_callback(name: str, doc: Document) -> None:
```

This makes it hard for users that wish to type annotate their callbacks to properly discern between different document types without having to rely on `# type: ignore` or `typing.cast`.

Using `DocumentNames` as a `StrEnum` and adding a `NamedDocument` union that packs the enum to its respective document type, we can do something like this:

```py
from event_model import NamedDocument, DocumentNames

def my_callback(doc: NamedDocument) -> None:
    match doc.name:
          case DocumentNames.start:
              reveal_type(doc.doc) # revealed type is RunStart
    if doc.name == Document.start:
        reveal_type(doc.doc) # revealed type is RunStart
```

Initially the attempt was to support *both* raw strings **and** StrEnum as a narrower for the document type. Apparently this is not possible. Our thought (mine and @evvaaaa) is that we should encourage users that wish for a more type-safe approach in this callback to use this narrowing policy via enumerators.

### DIFF
--- a/src/event_model/__init__.py
+++ b/src/event_model/__init__.py
@@ -5,7 +5,6 @@ import inspect
 import itertools
 import json
 import os
-import sys
 import threading
 import time as ttime
 import uuid
@@ -24,35 +23,12 @@ from typing import (
 import jsonschema
 import numpy
 
-# TODO: remove backport when Python 3.10 support is dropped
-if sys.version_info >= (3, 11):
-    from enum import StrEnum
-else:
-    from ._strenum_backport import StrEnum
-
-
-class DocumentNames(StrEnum):
-    stop = "stop"
-    start = "start"
-    descriptor = "descriptor"
-    event = "event"
-    datum = "datum"
-    resource = "resource"
-    event_page = "event_page"
-    datum_page = "datum_page"
-    stream_resource = "stream_resource"
-    stream_datum = "stream_datum"
-    bulk_datum = "bulk_datum"  # deprecated
-    bulk_events = "bulk_events"  # deprecated
-
-
-# Imported below DocumentNames because the generated documents package reads it
-# back off this module while this module is still initialising.
-from ._version import __version__  # noqa: E402
-from .documents.datum import Datum  # noqa: E402
-from .documents.datum_page import DatumPage  # noqa: E402
-from .documents.event import Event, PartialEvent  # noqa: E402
-from .documents.event_descriptor import (  # noqa: E402
+from ._version import __version__
+from .document_names import DocumentNames
+from .documents.datum import Datum
+from .documents.datum_page import DatumPage
+from .documents.event import Event, PartialEvent
+from .documents.event_descriptor import (
     Configuration,
     DataKey,
     Dtype,
@@ -61,9 +37,9 @@ from .documents.event_descriptor import (  # noqa: E402
     LimitsRange,
     PerObjectHint,
 )
-from .documents.event_page import EventPage, PartialEventPage  # noqa: E402
-from .documents.resource import PartialResource, Resource  # noqa: E402
-from .documents.run_start import (  # noqa: E402
+from .documents.event_page import EventPage, PartialEventPage
+from .documents.resource import PartialResource, Resource
+from .documents.run_start import (
     CalculatedEventProjection,
     Calculation,
     ConfigurationProjection,
@@ -73,9 +49,9 @@ from .documents.run_start import (  # noqa: E402
     RunStart,
     StaticProjection,
 )
-from .documents.run_stop import RunStop  # noqa: E402
-from .documents.stream_datum import StreamDatum, StreamRange  # noqa: E402
-from .documents.stream_resource import StreamResource  # noqa: E402
+from .documents.run_stop import RunStop
+from .documents.stream_datum import StreamDatum, StreamRange
+from .documents.stream_resource import StreamResource
 
 __all__ = [
     # Document types
--- a/src/event_model/__init__.py
+++ b/src/event_model/__init__.py
@@ -5,6 +5,7 @@ import inspect
 import itertools
 import json
 import os
+import sys
 import threading
 import time as ttime
 import uuid
@@ -13,7 +14,6 @@ import weakref
 from collections import defaultdict, deque
 from collections.abc import Callable, Generator, Iterable, Iterator
 from dataclasses import dataclass
-from enum import Enum
 from typing import (
     Any,
     Literal,
@@ -24,11 +24,35 @@ from typing import (
 import jsonschema
 import numpy
 
-from ._version import __version__
-from .documents.datum import Datum
-from .documents.datum_page import DatumPage
-from .documents.event import Event, PartialEvent
-from .documents.event_descriptor import (
+# TODO: remove backport when Python 3.10 support is dropped
+if sys.version_info >= (3, 11):
+    from enum import StrEnum
+else:
+    from ._strenum_backport import StrEnum
+
+
+class DocumentNames(StrEnum):
+    stop = "stop"
+    start = "start"
+    descriptor = "descriptor"
+    event = "event"
+    datum = "datum"
+    resource = "resource"
+    event_page = "event_page"
+    datum_page = "datum_page"
+    stream_resource = "stream_resource"
+    stream_datum = "stream_datum"
+    bulk_datum = "bulk_datum"  # deprecated
+    bulk_events = "bulk_events"  # deprecated
+
+
+# Imported below DocumentNames because the generated documents package reads it
+# back off this module while this module is still initialising.
+from ._version import __version__  # noqa: E402
+from .documents.datum import Datum  # noqa: E402
+from .documents.datum_page import DatumPage  # noqa: E402
+from .documents.event import Event, PartialEvent  # noqa: E402
+from .documents.event_descriptor import (  # noqa: E402
     Configuration,
     DataKey,
     Dtype,
@@ -37,9 +61,9 @@ from .documents.event_descriptor import (
     LimitsRange,
     PerObjectHint,
 )
-from .documents.event_page import EventPage, PartialEventPage
-from .documents.resource import PartialResource, Resource
-from .documents.run_start import (
+from .documents.event_page import EventPage, PartialEventPage  # noqa: E402
+from .documents.resource import PartialResource, Resource  # noqa: E402
+from .documents.run_start import (  # noqa: E402
     CalculatedEventProjection,
     Calculation,
     ConfigurationProjection,
@@ -49,9 +73,9 @@ from .documents.run_start import (
     RunStart,
     StaticProjection,
 )
-from .documents.run_stop import RunStop
-from .documents.stream_datum import StreamDatum, StreamRange
-from .documents.stream_resource import StreamResource
+from .documents.run_stop import RunStop  # noqa: E402
+from .documents.stream_datum import StreamDatum, StreamRange  # noqa: E402
+from .documents.stream_resource import StreamResource  # noqa: E402
 
 __all__ = [
     # Document types
@@ -89,21 +113,6 @@ __all__ = [
     "compose_run",
     "__version__",
 ]
-
-
-class DocumentNames(Enum):
-    stop = "stop"
-    start = "start"
-    descriptor = "descriptor"
-    event = "event"
-    datum = "datum"
-    resource = "resource"
-    event_page = "event_page"
-    datum_page = "datum_page"
-    stream_resource = "stream_resource"
-    stream_datum = "stream_datum"
-    bulk_datum = "bulk_datum"  # deprecated
-    bulk_events = "bulk_events"  # deprecated
 
 
 class DocumentRouter:
--- a/src/event_model/documents/__init__.py
+++ b/src/event_model/documents/__init__.py
@@ -1,5 +1,13 @@
 # generated in `event_model/generate`
 
+from typing import Generic, Literal, TypeVar
+
+# TODO: import from typing when Python 3.10 support is dropped, where a
+# NamedTuple cannot yet be generic.
+from typing_extensions import NamedTuple
+
+from event_model import DocumentNames
+
 from .datum import *  # noqa: F403
 from .datum_page import *  # noqa: F403
 from .event import *  # noqa: F403
@@ -10,6 +18,20 @@ from .run_start import *  # noqa: F403
 from .run_stop import *  # noqa: F403
 from .stream_datum import *  # noqa: F403
 from .stream_resource import *  # noqa: F403
+
+NameT = TypeVar("NameT")
+DocT = TypeVar("DocT")
+
+
+class DocWrapper(Generic[NameT, DocT], NamedTuple):
+    """Named document wrapper."""
+
+    name: NameT
+    """The name each document is emitted under."""
+
+    doc: DocT
+    """The document itself."""
+
 
 DocumentType = (
     type[Datum]  # noqa: F405
@@ -35,6 +57,30 @@ Document = (
     | RunStop  # noqa: F405
     | StreamDatum  # noqa: F405
     | StreamResource  # noqa: F405
+)
+
+NamedDatum = DocWrapper[Literal[DocumentNames.datum], Datum]  # noqa: F405
+NamedDatumPage = DocWrapper[Literal[DocumentNames.datum_page], DatumPage]  # noqa: F405
+NamedEvent = DocWrapper[Literal[DocumentNames.event], Event]  # noqa: F405
+NamedEventDescriptor = DocWrapper[Literal[DocumentNames.descriptor], EventDescriptor]  # noqa: F405
+NamedEventPage = DocWrapper[Literal[DocumentNames.event_page], EventPage]  # noqa: F405
+NamedResource = DocWrapper[Literal[DocumentNames.resource], Resource]  # noqa: F405
+NamedRunStart = DocWrapper[Literal[DocumentNames.start], RunStart]  # noqa: F405
+NamedRunStop = DocWrapper[Literal[DocumentNames.stop], RunStop]  # noqa: F405
+NamedStreamDatum = DocWrapper[Literal[DocumentNames.stream_datum], StreamDatum]  # noqa: F405
+NamedStreamResource = DocWrapper[Literal[DocumentNames.stream_resource], StreamResource]  # noqa: F405
+
+NamedDocument = (
+    NamedDatum  # noqa: F405
+    | NamedDatumPage  # noqa: F405
+    | NamedEvent  # noqa: F405
+    | NamedEventDescriptor  # noqa: F405
+    | NamedEventPage  # noqa: F405
+    | NamedResource  # noqa: F405
+    | NamedRunStart  # noqa: F405
+    | NamedRunStop  # noqa: F405
+    | NamedStreamDatum  # noqa: F405
+    | NamedStreamResource  # noqa: F405
 )
 
 ALL_DOCUMENTS: tuple[DocumentType, ...] = (
--- a/src/event_model/documents/__init__.py
+++ b/src/event_model/documents/__init__.py
@@ -6,7 +6,7 @@ from typing import Generic, Literal, TypeVar
 # NamedTuple cannot yet be generic.
 from typing_extensions import NamedTuple
 
-from event_model import DocumentNames
+from event_model.document_names import DocumentNames
 
 from .datum import *  # noqa: F403
 from .datum_page import *  # noqa: F403
--- a/src/event_model/generate/create_documents.py
+++ b/src/event_model/generate/create_documents.py
@@ -227,7 +227,7 @@ from typing import Generic, Literal, TypeVar
 # NamedTuple cannot yet be generic.
 from typing_extensions import NamedTuple
 
-from event_model import DocumentNames
+from event_model.document_names import DocumentNames
 
 {imports}
 
--- a/src/event_model/generate/create_documents.py
+++ b/src/event_model/generate/create_documents.py
@@ -11,9 +11,25 @@ from pydantic.json_schema import GenerateJsonSchema
 
 from event_model.basemodels import ALL_BASEMODELS
 
+#: Path to JSON schema files.
 JSONSCHEMA = Path(__file__).parent.parent / "schemas"
+
+#: Path to TypedDict documents.
 TYPEDDICTS = Path(__file__).parent.parent / "documents"
-TEMPLATES = Path(__file__).parent / "templates"
+
+#: Name each document is emitted under, keyed by module name.
+DOCUMENT_NAMES: dict[str, str] = {
+    "datum": "datum",
+    "datum_page": "datum_page",
+    "event": "event",
+    "event_descriptor": "descriptor",
+    "event_page": "event_page",
+    "resource": "resource",
+    "run_start": "start",
+    "run_stop": "stop",
+    "stream_datum": "stream_datum",
+    "stream_resource": "stream_resource",
+}
 
 
 class SnakeCaseTitleField(GenerateJsonSchema):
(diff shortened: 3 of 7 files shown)


jacopoabramo:
@evvaaaa @coretl

src/event_model/tests/test_named_documents.py
```diff
@@ -0,0 +1,186 @@
...
+        validator.validate(doc)
+
+    assert {name for name, _ in composed} == set(get_args(DocumentName))
+
+
+def test_callback_discriminates_documents_by_name(tmp_path):
+    """A callback taking one ``(name, doc)`` tuple can dispatch on the name."""
+    seen = {}
+
+    def callback(named_document: NamedDocument) -> None:
```
evvaaaa: It's cool that type checkers will now understand this.

``` python
def datum(doc: Datum): ...
def event(doc: Event): ...

def foo(named_doc: NamedDocument):
    if named_doc[0] == "datum":
        datum(named_doc[1])
        event(named_doc[1])  # Fails type hinting
    match named_doc:
        case ("datum", doc):
            datum(doc)
            event(doc)  # Fails type hinting
```

I was playing around with a way of making variables for these strings work, e.g

``` python
DATUM_NAME: Literal["datum"] = "datum"
```

but no dice unfortunately.

evvaaaa: Actually, maybe it would be nice to add

``` python
DATUM_NAME = "datum"
DATUM_PAGE_NAME = "datum_page"
EVENT_NAME = "event"
EVENT_PAGE_NAME = "event_page"
DESCRIPTOR_NAME = "descriptor"
RESOURCE_NAME = "resource"
START_NAME = "start"
STOP_NAME = "stop"
STREAM_DATUM_NAME = "stream_datum"
STREAM_RESOURCE_NAME = "stream_resource"
```

to `event_model.documents.__init__.py`:

``` python
def foo(named_doc: NamedDocument):
    if named_doc[0] == DATUM_NAME:
        datum(named_doc[1])  # Type checker understands this is a datum
        event(named_doc[1])  # Fails type hinting
    match named_doc:
        case (n, d) if n == EVENT_NAME:
            datum(d) # Fails type hinting
            event(d) # Type checker thinks this could be any document
```

jacopoabramo: Ok, so... both your examples (using `if named_doc[0]` or `match named_doc`) work at runtime.

At type checking time (with mypy), the latter fails as it is. To explicitly make it work, you have to use `documents.<document_name_variable>` in your code, which I find incredibly annoying.

You can test this locally with this failing example if you fork this branch:

```py
def test_failing_match(tmp_path):
    seen = {}

    def callback(named_document: NamedDocument) -> None:
        match named_document:
            case (name, doc) if name == START_NAME:
                seen[START_NAME] = doc["uid"]
            case (name, doc) if name == DESCRIPTOR_NAME:
                seen[DESCRIPTOR_NAME] = sorted(doc["data_keys"])
            case (name, doc) if name == RESOURCE_NAME:
                seen[RESOURCE_NAME] = doc["resource_path"]
            case (name, doc) if name == STREAM_RESOURCE_NAME:
                seen[STREAM_RESOURCE_NAME] = doc["uri"]
            case (name, doc) if name == DATUM_NAME:
                seen[DATUM_NAME] = doc["datum_id"]
            case (name, doc) if name == DATUM_PAGE_NAME:
                seen[DATUM_PAGE_NAME] = doc["datum_id"]
            case (name, doc) if name == EVENT_NAME:
                seen[EVENT_NAME] = doc["seq_num"]
            case (name, doc) if name == EVENT_PAGE_NAME :
                seen[EVENT_PAGE_NAME] = doc["seq_num"]
            case (name, doc) if name == STREAM_DATUM_NAME:
                seen[STREAM_DATUM_NAME] = doc["seq_nums"]
            case (name, doc) if name == STOP_NAME:
                seen[STOP_NAME] = doc["exit_status"]
            case _:
                raise AssertionError(f"{named_document[0]} matched no case")

    composed = compose_all_documents(tmp_path)
    for named_document in composed:
        callback(named_document)

    assert_every_document_was_routed(seen, composed)
```

jacopoabramo: Rationale from Claude:

> PEP 634 gives match its own grammar in which a bare name is always a capture pattern, decided by syntax alone, before any name resolution. There's no marker meaning "treat this bare name as a value" because that would make a pattern's meaning depend on what happens to be in scope at that point: adding an unrelated import could silently turn a binding into a comparison. A dot is the disambiguator they picked, so a value pattern must be mod.NAME, Class.NAME, or Enum.MEMBER. Alternative spellings like .NAME or ==NAME were considered and rejected in PEP 635.

Basically match-case requires that the casing is expressed via a dot syntax of some kind. Given this, instead of plain variable names, how about making it as an enum class?

```py
from enum import Enum

# I'm open to suggestions on the class name
class DocumentNameEnum(Enum):
    START = "start"
    ... # and so on and so forth
```

evvaaaa: We actually already have an enum class, was going to comment that 

https://github.com/bluesky/event-model/blob/017f073df6572dae426921cb31e918557a869f0e/src/event_model/__init__.py#L94-L106

should use the new literals in the definition.

> how about making it as an enum class?

I was thinking about this but then considered that it would only make sense if we changed the run engine to emit the enum instead of the string, otherwise the user would have to always add

``` python
name, doc = DocumentName(named_doc[0]), named_doc[1]
```

But on second though, with a `StrEnum` we could have our cake and eat it too.

``` python
class DocumentName(StrEnum): # Renamed from document name
   STOP = "stop"
   START = "start"
   DESCRIPTOR = "descriptor"
   # and so on
```

Then bluesky callbacks are `Callable[[DocumentName, Document], Any]`, and 

### previous checks with equality don't have to be changed, e.g 

``` python
def foo(named_doc):
    if named_doc[0] == "datum":
        datum(named_doc[1])
```

still passes because `DocumentName.DATUM == "datum"`, and 

### we have even nicer typed match statements

``` python
# Excerpt for brevity
class DocumentName(StrEnum):
    STOP = "stop"
    START = "start"
    DESCRIPTOR = "descriptor"
    EVENT = "event"


NamedEvent = tuple[Literal[DocumentName.EVENT], Event]  # noqa: F405
NamedEventDescriptor = tuple[Literal[DocumentName.DESCRIPTOR], EventDescriptor]  # noqa: F405
NamedRunStart = tuple[Literal[DocumentName.START], RunStart]  # noqa: F405
NamedRunStop = tuple[Literal[DocumentName.STOP], RunStop]  # noqa: F405


NamedDocument = (
    NamedRunStart
    | NamedEvent  # noqa: F405
    | NamedEventDescriptor  # noqa: F405
    | NamedRunStop  # noqa: F405
)


def start(doc: RunStart):
    print("datum called")


def event(doc: Event):
    print("event called")


def callback(named_doc: NamedDocument):
    match named_doc:
        case (DocumentName.START, d):
            start(d)
            event(d)  # Fails type hinting
        case (DocumentName.EVENT, d):
            start(d)  # Fails type hinting
            event(d)
```

jacopoabramo: @evvaaaa the only problem is that `StrEnum` is `python>=3.11` only.

What we *could* do is vendor [backports.strenum](https://pypi.org/project/backports.strenum/), and then use the default `StrEnum` once we drop support for python 3.10 (it'll go EOL in october, not so distant in the future).

A more practical approach would be to simply add `backports.strenum` as a dependency. I'm suggesting vendoring because I did this work in the past for napari and they have a more complex management of dependencies than event-model has.

What do you think? Can we afford an extra dependency (even if for just a release)? Or is vendoring fine? We can't drop Python 3.10 until Bluesky does it too

EDIT: my feeling is to vendor it, just because it doesn't justify adding a dependency out of nowhere. I still would really like to use StrEnum, it would open some opportunities for cleaning up some things I remember seeing in the bluesky code that I didn't like

evvaaaa: Go for it! It's tiny and does the same `(Enum, str)` inheritance we used to do in ophyd async.

https://github.com/clbarnes/backports.strenum/blob/18c42ffd77615441eff10bc2f4e3d98d44722d72/src/backports/strenum/strenum.py#L8

``` python
# event_model/strenum.py
from enum import Enum
from typing import TypeVar
import sys

if sys.version_info >= (3, 11):
    from enum import StrEnum
else:
    _S = TypeVar("_S", bound="StrEnum")

    class StrEnum(str, Enum):
        def __new__(cls: type[_S], *values: str) -> _S:
            if len(values) > 3:
                raise TypeError("too many arguments for str(): %r" % (values,))
            if len(values) == 1:
                # it must be a string
                if not isinstance(values[0], str):
                    raise TypeError("%r is not a string" % (values[0],))
            if len(values) >= 2:
                # check that encoding argument is a string
                if not isinstance(values[1], str):
                    raise TypeError("encoding must be a string, not %r" % (values[1],))
            if len(values) == 3:
                # check that errors argument is a string
                if not isinstance(values[2], str):
                    raise TypeError("errors must be a string, not %r" % (values[2]))
            value = str(*values)
            member = str.__new__(cls, value)
            member._value_ = value
            return member

        __str__ = str.__str__

        @staticmethod
        def _generate_next_value_(
            name: str, start: int, count: int, last_values: list[str]
        ) -> str:
            return name.lower()


__all__ = ["StrEnum"]
```

jacopoabramo:
I completely forgot about the whole discussion we had in the issue.

I revised the implementation completely. It now aligns better with the strategy @coretl suggested

I force-pushed a single commit to simplify reviewing (I hope at least).

I'm happy with the tests so I'll mark this ready for review.

src/event_model/documents/__init__.py
```diff
@@ -37,6 +53,30 @@
...
+NamedEvent = DocWrapper[Literal["event"], Event]  # noqa: F405
+NamedEventDescriptor = DocWrapper[Literal["descriptor"], EventDescriptor]  # noqa: F405
+NamedEventPage = DocWrapper[Literal["event_page"], EventPage]  # noqa: F405
+NamedResource = DocWrapper[Literal["resource"], Resource]  # noqa: F405
+NamedRunStart = DocWrapper[Literal["start"], RunStart]  # noqa: F405
+NamedRunStop = DocWrapper[Literal["stop"], RunStop]  # noqa: F405
+NamedStreamDatum = DocWrapper[Literal["stream_datum"], StreamDatum]  # noqa: F405
+NamedStreamResource = DocWrapper[Literal["stream_resource"], StreamResource]  # noqa: F405
+
+NamedDocument = (
```
evvaaaa, line 73: We should change bluesky typing at some point to explicitly give `NamedDocument`.

Not necessary in the short term.

src/event_model/generate/create_documents.py
```diff
@@ -205,43 +224,82 @@ def generate_jsonschema(
...
-        f"{snake_to_title(document_name)}" for document_name in document_names
+        f"{snake_to_title(module_name)}" for module_name in module_names
     ]
 
     init_py_imports = "\n".join(
         sorted(
-            [
-                f"from .{document_name} import *  # noqa: F403"
-                for document_name in document_names
+            [f"from .{HELPER} import DocWrapper"]
```
evvaaaa: I think it'd be nicer to:

1) Move `StrEnum` to `event_model/strenum_3_10_backport.py`
2) Move `DocWrapper` to `event_model/documents/__init__.py` (the top of the static string in create_documents.py) directly - in python3.10 you can import `TypedDict` from both `typing` and `typing_extensions` I believe.
3) delete `helper.py`

evvaaaa (approved):
Early approval but one relevant comment

src/event_model/documents/__init__.py
jacopoabramo: Sorry I'm not following, you're referring to the bluesky codebase? Where exactly?

src/event_model/generate/create_documents.py
jacopoabramo: > in python3.10 you can import TypedDict from both typing and typing_extensions I believe

~Yeah... Claude hallucination, my fault for not picking it up. It was convinced subclassing TypedDict with a generic was not possible; turns out you just need to swap the inheritance order.~

Actually it was right, I re-run the tests... this confuses me slightly, because [Reading](https://github.com/bluesky/bluesky/blob/6f839a1a2d35ae15c9ffd291e4f7723c66ba9353/src/bluesky/protocols.py#L36) does the same pattern.

The difference is that it's not used at runtime, I guess?

Ah, no, actually Reading uses TypedDict from typing_extensions so I guess it's the same problem. I don't think we can go around that.

I'll apply the changes.

src/event_model/documents/__init__.py
evvaaaa: I mean editing the [dispatcher](https://github.com/bluesky/bluesky/blob/6f839a1a2d35ae15c9ffd291e4f7723c66ba9353/src/bluesky/run_engine.py#L2675) but it can probably wait.

It's maybe worth a little check that bluesky main still passes type hinting and tests with this branch of event model.

jacopoabramo: Most of the run_engine module is untyped, dispatcher in particular, so I'm not *too* worried, but I'll try locally to see if it needs adjustments (either tonight or tomorrow)

src/event_model/documents/__init__.py
```diff
@@ -11,6 +17,33 @@
...
+    "datum_page",
+    "event",
+    "descriptor",
+    "event_page",
+    "resource",
+    "start",
+    "stop",
+    "stream_datum",
+    "stream_resource",
+]
```
jacopoabramo: I'm just realizing this is probably a leftover from the previous approach, ok to remove it?

evvaaaa: Yes please do.

src/event_model/documents/__init__.py
```diff
@@ -37,6 +70,30 @@
...
+NamedDatum = DocWrapper[Literal["datum"], Datum]  # noqa: F405
+NamedDatumPage = DocWrapper[Literal["datum_page"], DatumPage]  # noqa: F405
+NamedEvent = DocWrapper[Literal["event"], Event]  # noqa: F405
+NamedEventDescriptor = DocWrapper[Literal["descriptor"], EventDescriptor]  # noqa: F405
+NamedEventPage = DocWrapper[Literal["event_page"], EventPage]  # noqa: F405
+NamedResource = DocWrapper[Literal["resource"], Resource]  # noqa: F405
+NamedRunStart = DocWrapper[Literal["start"], RunStart]  # noqa: F405
+NamedRunStop = DocWrapper[Literal["stop"], RunStop]  # noqa: F405
+NamedStreamDatum = DocWrapper[Literal["stream_datum"], StreamDatum]  # noqa: F405
+NamedStreamResource = DocWrapper[Literal["stream_resource"], StreamResource]  # noqa: F405
```
evvaaaa: Might be a bit nicer to store `DocumentNames` in this file here and use e.g `Literal[DocumentNames.stream_resource]`?

jacopoabramo: Unfortunately it would break the behavior entirely. Or rather: my expectation is that user prefer the raw string values for narrowing. When replacing with `Literal[DocumentNames.<document>]` we would lose the ability to check the raw strings, and viceversa.

So currently `DocumentNames` will only be helpful for better typing/ergonomics in the bluesky dispatcher in the future, but not for narrowing documents.

Apparently this breaks both with mypy and pyright; i'll try to reproduce in a playground.

src/event_model/documents/__init__.py
jacopoabramo: Done

src/event_model/documents/__init__.py
jacopoabramo: [mypy playground](https://mypy-play.net/?mypy=latest&python=3.11&gist=f892ca306a1047c396e3bc9acc12f411)
[pyright playground](https://pyright-play.net/?code=GYJw9gtgBApgdgV2gSwgBzCALlAooiAKFEiiwE81k4BzKVDbKAcXhhGQGMAaKAGWRZ2AQwA2vACqUYAEwAiXLJOkA1YSEKbOo4QGddUOWE5J4WAHLCIMXQApdWEL3xIAlAC5CUb1AfqcALxQAER%2B2MFePjLCWEhQQcHRsRARWjr6UABKCHAAylj%2BtlJosgqcWB6R3gjIMu6%2BjpqE2noGcjFIRdLyipU%2BUElIAPq19Q4ampbWEvFQxTBqILbBUzASwa6ERpwzQfOLy9vrm2mthsYA6iDCaCVLrHDsXADaqxK8RwC6yiU95X0%2BOBWGD1N5VAbGepHJr5EAAaRg5Fks22Vxud2eAiE11Ez1CBXC3yyOXy-k%2BUAAPudOGjbuxMYIRLjEh0UkT2slPpoXBAEUiZLNbODUdc6SAGdixM9tqY4BZgboAHRhLBE7J5Amq8FUkXo%2BlYpnS4yy%2BXWJWDCDs1lck6EGQwYANEBDADWiNkQwA7oIABZDcbUGi2OB1KCwvmyVxQAC0AD4oOYwI9PP1kI6Q3igdZguSAgkVcEU-0fCAYAA3GBiIYUErBmR4mTGHNRqAAYmJGv8TXtjvGrvdMi9vqG1ggACN2HX6uGB1G4wmkyDwWmoBnglmYDn4kEZdY5aslSqi8WoKWK1WazA6w2m58W%2B31aSmDrWd2HbACP3%2BUOsH6A7QpzwAgIxkOd40TZNl3Tet12BLc8xCAtj2LM9K1EatpGvRJb3vDsnxwF9kjfR14GGN1v29X8RxgcdJxDeoeRAsCF0g1NoMzODcx3Y091NGxlU1ZD%2BlQi9MLXRtOGbbwHxJTUgA)

basically the behaviors are mutually exclusive. considering the divergence between mypy and pyright (for the latter, a few cases are unreachable) it's either a bug, an intended behavior by both, or something lacking in the spec.

for now I would prefer keeping literal strings, just because it's the de facto standard approach of most users

evvaaaa: Ah sorry I misread right before leaving work and missed that `DocWrapper` was `TypedDict` 🙏 

# 1

I'm wondering if why exactly we're deciding on a new typeddict here instead of a tuple? A tuple fits in with arguments callbacks currently take, and I think a match statement on `(name, document)` is a little nicer.

The perspective in the issue was that we could either:

1. Enable typing for the current (name, document) tuple
2. Add a new field `"document_type"` to each document to allow the match on the index which is being done in your branches `DocWrapper`.

# 2 (relevant whether we use a tuple or a new dict)

Example for precision:

``` python
NamedEvent = tuple[Literal[DocumentNames.event], Event]
NamedRunStart = tuple[Literal[DocumentNames.start], RunStart]


NamedDocument = (
    NamedRunStart | NamedEvent  # noqa: F405
)

def case1(named_doc: NamedDocument):
    match named_doc:
        case (DocumentNames.start, d):
            start(d)
            # event(d)  # Fails type hinting
    if named_doc[0] == DocumentNames.start:
        start(named_doc[1])
        # event(named_doc[1]) # Fails type hinting


def case2(named_doc: NamedDocument):
    match named_doc[0]:
        case "start":
            # Type hinting thinks code is unreachable but works at run time
            start(named_doc[1])
    if named_doc[0] == "start":
        # Type hinting thinks code is unreachable but works at run time
        start(named_doc[1])
```

> Unfortunately it would break the behavior entirely. Or rather: my expectation is that user prefer the raw string values for narrowing. When replacing with Literal[DocumentNames.<document>] we would lose the ability to check the raw strings, and viceversa.

I'm inclined to say that if the user cares about typing then we should force them to use `case1`, and if they don't then their code will still run. Checks with raw strings in callbacks are already either:

1. Untyped
2. Coerced with type ignores or `typing.cast`s. 

In both cases they won't notice a difference, but if 2. changes their code to enable proper typing then they can also start using `DocumentNames.` because it's nicer.

If we decide on a new `document_type` field to each document then I'm pretty sure we can still type hint it with `Literal[DocumentNames.x]` since it will serialize to string.

jacopoabramo: > I'm wondering if why exactly we're deciding on a new typeddict here instead of a tuple?

I reviewed the issue, and I mixed up a lot of the discussion. The final consensus was to have indeed a `NamedTuple` and not a `TypedDict` as I'm doing now. So the approach would be this one:

```py
from typing import TypedDict, Literal, TypeVar, Generic, NamedTuple, reveal_type

class DocA(TypedDict):
    foo: str
    bar: int

class DocB(TypedDict):
    foo: str
    bat: float

NameT = TypeVar("NameT")
DocT = TypeVar("DocT")

class NamedDocument(NamedTuple, Generic[NameT, DocT]):
    name: NameT
    doc: DocT
    

NamedDocumentUnion = \
    NamedDocument[Literal["doc_a"], DocA] | \
    NamedDocument[Literal["doc_b"], DocB]


def my_callback(named_document: NamedDocumentUnion):
    match named_document:
        case "doc_a", doc:
            num = doc["bar"]
        case "doc_b", doc:
            num = doc["bat"]
    reveal_type(num)
```

which as of now works with both mypy and pyright. So this:

> 1. Enable typing for the current (name, document) tuple

is what I actually needed to aim for, but I lost track of that in the discussion. I'll adjust it. Sorry!

> I'm inclined to say that if the user cares about typing then we should force them to use case1, and if they don't then their code will still run.

I'm sold on this. I'll adjust it to enforce using document names. Hopefully it'll also help in bluesky in the future.

src/event_model/__init__.py
```diff
@@ -24,11 +24,35 @@
...
+    descriptor = "descriptor"
+    event = "event"
+    datum = "datum"
+    resource = "resource"
+    event_page = "event_page"
+    datum_page = "datum_page"
+    stream_resource = "stream_resource"
+    stream_datum = "stream_datum"
+    bulk_datum = "bulk_datum"  # deprecated
+    bulk_events = "bulk_events"  # deprecated
```
jacopoabramo: This is, frankly, horrible. I'm putting it here temporarely just to show the circular import problem.

Best I can think of is to generate DocumentNames as well. Not very pretty but if we want to keep the documents module clean it's the only think I can think of.

jacopoabramo (commented):
We're hopefully getting there, there's just a slight hiccup to address yet.

src/event_model/__init__.py
evvaaaa: Hopefully it won't be as much of a problem after

https://github.com/bluesky/bluesky/issues/2036#issuecomment-5043220724

Maybe this should live in bluesky/document_names.py?

It doesn't seem specific to `event_model.documents` or jsonschema `event_model.__init__.py`

jacopoabramo:
@evvaaaa hopefully this is the last review I ask you to do @_@ sorry, I thought it was gonna be easier

evvaaaa (approved):
LGTM!

